PR #224, opened 2019-09-25
Add MaPopupModal component

Samitier:
Adds the popup component copied as-is from [this Maitai component](https://github.com/holaluz/maitai/tree/master/src/components/PopupModal), including some testing and docs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8760,9 +8760,14 @@
       }
     },
     "date-fns": {
-      "version": "1.30.1",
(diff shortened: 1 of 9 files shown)


Samitier:
closed cuz this is older than covid

charlybrown03:
> closed cuz this is older than covid

image

Maybe yes, but maybe covid can be older (but not known yet), as it was "tagged" as a new disease on Nov 17th 😛 .

Happy to see all these movements in Margarita.

Regards!! 🤗

PR #249, opened 2020-03-17
[DO NOT MERGE] Add CMB module

Samitier:
This is an "educative" PR not meant to be merged.

Please read https://github.com/holaluz/blog/pull/155 first.

This PR ports the `dry-martini` CMB module to margarita to be compiled as a standalone component. In order to do that, some changes had to be done:

- Import paths changed (obviously)
- Privacy policy link does not open a modal (to remove even more dependencies). Now its just a link to the privacy policy page.
- Scoped i18n. The `dry-martini` component used global locales and I had to scope them.
- The `dry-martini` component used global css and I also had to scope it.
- Also, it used the `nuxt-i18n` dependency, which we can't use in margarita. I just did a basic plugin that pushes events to dataLayer by hand.
- Finally, I wrapped all the component in a new instance of vue, so it can be rendered only executing a function (which makes things way cleaner in holaluz/blog).

What I get from all this:
- Please, pleeeease let's scope all the css and locales. Moving the component would have been trivial if we had done that.
- I'm now interested in vue plugins as a way to replicate dependency injection. I am thinking of, for example, make our apis vue plugins, so we can inject a mock api service in all of our app with just one line of code. Something like this:
```js
Vue.prototype.$fooApi = node.env.ENV === 'development' ? mockFooService : fooService
```
- Let's do a molecule repo! Same build process as margarita, same idea in mind as this PR. Do you agree?

Thanks for reading this much!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1036,6 +1036,18 @@
         "debug": "^3.0.1",
         "openurl": "1.1.1",
         "yargs": "6.6.0"
+      },
+      "dependencies": {
+        "axios": {
+          "version": "0.19.0",
+          "resolved": "https://registry.npmjs.org/axios/-/axios-0.19.0.tgz",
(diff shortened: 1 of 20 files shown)


Samitier:
Closed as it has already been implemented in kalimotxo

PR #265, opened 2020-08-13
Add beta release and move CI/CD to GHActions

charlybrown03:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,20 @@
+name: Publish NPM package
```
afontcu, line 1: Should we add a testing step here, just in case? So we make sure we don't publish a broken release. Maybe changing `prepublishOnly` script to run `npm it` ([docs](https://docs.npmjs.com/cli/install-test.html)) instead of `npm i` would do the trick

PR #269, opened 2020-08-20
fix: v-markdown directive for SSR

Samitier:
We have a problem with `v-markdown` directive 🔥

Looks like nuxt doesn't render directives server-side [unless we specify stuff in `nuxt.config.js`](https://github.com/nuxt/nuxt.js/issues/898). This means that all content set with `v-markdown` is not in the SSR page and it makes some popping when the app hydrates because the content [appears after the page loads](https://www.holaluz.com/autoconsumo/). We haven't had any trouble with SEO, everything indexes correctly, but I guess its better to not risk it.

We should add this to every `nuxt.config` of the SSR projects that use `v-markdown`:

```js
import { markdown } from '@holaluz/margarita'

[...]

render: {
    bundleRenderer: {
      directives: {
        markdown(vnode, directiveMeta) {
          vnode.data.domProps = {
            ...vnode.data.domProps,
            childNodes: [],
          }
          markdown(vnode.data.domProps, directiveMeta)
        },
      },
    },
  },
```
In this PR I added the SSR implementation of `v-markdown` directly in margarita so we can do something like this:

```js
import { markdownSSR } from '@holaluz/margarita'

[...]

render: {
    bundleRenderer: {
      directives: {
         markdown: markdownSSR
      }
    }
}
```

### DIFF
--- a/src/index.spec.js
+++ b/src/index.spec.js
@@ -4,12 +4,14 @@ import isVueComponent from 'is-vue-component'
 
 import * as Margarita from './index'
 
-// Get the default export...
+// Get the default export
 const defaultExport = Margarita.default
 
-// ...and remove it from the object, so we have the list of imported comps.
-// eslint-disable-next-line no-import-assign
-delete Margarita.default
+// Remove default export and directives, so we have the list of imported comps.
+const margarita = { ...Margarita }
+delete margarita.default
+delete margarita.markdown
+delete margarita.markdownSSR
 
 test('exposes an install function', () => {
   expect(defaultExport).toHaveProperty('install')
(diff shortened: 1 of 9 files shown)


afontcu:
I vote for the 3rd solution, too! However I'd not introduce if statements in unit tests 😬

Samitier:
> I vote for the 3rd solution, too! However I'd not introduce if statements in unit tests 😬

Added markdown SSR & fixed the tests

src/index.spec.js
```diff
@@ -4,12 +4,14 @@ import isVueComponent from 'is-vue-component'
...
 import * as Margarita from './index'
 
-// Get the default export...
+// Get the default export
 const defaultExport = Margarita.default
 
-// ...and remove it from the object, so we have the list of imported comps.
-// eslint-disable-next-line no-import-assign
-delete Margarita.default
+// Remove default export and directives, so we have the list of imported comps.
```
afontcu, line 10: neat 👌

afontcu:
Could we add the config snippet in the codebase, or in the story, or somewhere? To make it easy for everyone to find the right Nuxt config for markdown SSR 👌

Samitier:
> Could we add the config snippet in the codebase, or in the story, or somewhere? To make it easy for everyone to find the right Nuxt config for markdown SSR 👌

image

I added this. What do you think?

afontcu (approved):
neat 👌

PR #270, opened 2020-08-26
Update colors to match Design System

### LINKED ISSUE #254: Adapt Colors to the new Design System
## What

Refactor all Margarita's colors so that they comply with thew ones in the design system:

image

## How (lets discuss it)
 
My proposal: just create a `colors.scss` variable file with the colors and gradients that appear in the design system. To use them, we will have to link `margarita.scss` in style resources and just do:

```css
color: $hot-pink;
```

**Other ideas**: utility classes, same api as now (`get-color`), use the function of the color as color name ( main-primary instead of hot-pink), both utility classes and scss vars, etc.

## Tasks

- [ ] Decide the "how".
- [ ] Implement new approach.
- [ ] Implement figma color page in Storybook.
- [ ] Fix breaking changes (see below).

## Breaking changes

No matter how we do this, we will have to refactor hundreds of lines of code. I think they will all be "trivial" changes, until we find a color that it's not in the new Design System. In that case, I suggest creating it as a global scss variable in the project where it's being used.

afontcu:
Closes #254. This is obviously a huge breaking change 😄 

---

The proposal here is to use color maps to rely on way less variable names. This will make color usage way more simple (no more "Is this primrose? tundora? pippin" but rather "this is pink dark").

To do so, we leverage Sass' maps (a.k.a. arrays 🤷) to handle hues and variations.

New colors map: https://github.com/holaluz/margarita/blob/colors/src/scss/variables/_colors.scss

New  get-color function: https://github.com/holaluz/margarita/blob/colors/src/scss/functions/_colors.scss#L7

Result:

image

(_disclaimer: color ordering in this storybook screenshot is soooo poor, but I didn't want to spend much time there. Since colors are properly organized in Figma, we can easily check them out there_).

---

Out of Scope:

- gradients
- shadows
- fixing icons with Hex colors
- improving storybook's page on colors

---

- [x] Implement colors from Figma (currently missing "100% verde" colors)
- [x] Add new #F9F8F3 gray.
- [x] (_related to above_) Should we merge ink/sky colors into some "gray" map?
- [x] Agree on color naming. I kinda made up all names xD
- [x] Make sure old-new color equivalences are OK

The current equivalence of old-new colors is:

| old              | new          |
|------------------|--------------|
| dark-hot-pink    | pink         |
| classic-rose     | pink light   |
| pippin           | red light    |
| crimson          | red          |
| humming-bird     | blue light   |
| solitude     | blue light   |
| calypso          | blue         |
| picton-blue      | blue         |
| primrose         | green light  |
| japanese-laurel  | green        |
| peach-cream      | yellow light |
| brown            | yellow       |
| school-bus-yellow | yellow dark  |
| selective-yellow | yellow dark  |
| tango            | orange       |
| buttercup        | orange       |
| mine-shaft          | gray darker    |
| tundora          | gray dark    |
| emperor          | gray dark    |
| boulder          | gray          |
| scorpion         | gray          |
| gray         | gray          |
| alto             | gray light    |
| concrete         | brown light    |
| wild-sand        | brown ligth        |

### DIFF
--- a/src/assets/icons/warning.svg
+++ b/src/assets/icons/warning.svg
@@ -1,3 +1,3 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 22 19" fill="#945d00">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 22 19" fill="#f5d154">
 <path d="M0,19H22L11,0Zm12-3H10V14h2Zm0-4H10V8h2Z"/>
 </svg>
--- a/src/components/MaAlert/MaAlert.scss
+++ b/src/components/MaAlert/MaAlert.scss
@@ -1,6 +1,6 @@
 .alert-banner {
   display: flex;
-  color: get-color(ink);
+  color: get-color(gray, darkest);
 
   &--small {
     padding: rem(4px);
--- a/src/components/MaAlert/MaAlert.scss
+++ b/src/components/MaAlert/MaAlert.scss
@@ -1,5 +1,6 @@
 .alert-banner {
   display: flex;
+  color: get-color(ink);
 
   &--small {
     padding: rem(4px);
@@ -61,7 +62,6 @@
   }
 
   &__text {
-    @include font-weight(medium);
     margin: 0;
     width: 100%;
   }
@@ -75,22 +75,22 @@
   }
 
   &--error {
+    border: 1px solid get-color(red);
     background-color: get-color(red, light);
-    color: get-color(red);
   }
 
   &--info {
+    border: 1px solid get-color(blue);
     background-color: get-color(blue, light);
-    color: get-color(blue);
   }
 
   &--success {
+    border: 1px solid get-color(green);
     background-color: get-color(green, light);
-    color: get-color(green);
   }
 
   &--warning {
+    border: 1px solid get-color(yellow);
     background-color: get-color(yellow, light);
-    color: get-color(yellow);
   }
 }
--- a/src/components/MaAlert/MaAlert.scss
+++ b/src/components/MaAlert/MaAlert.scss
@@ -75,22 +75,22 @@
   }
 
   &--error {
-    background-color: get-color('pippin');
-    color: get-color('crimson');
+    background-color: get-color(red, light);
+    color: get-color(red);
   }
 
   &--info {
-    background-color: get-color('humming-bird');
-    color: get-color('calypso');
+    background-color: get-color(blue, light);
+    color: get-color(blue);
   }
 
   &--success {
-    background-color: get-color('primrose');
-    color: get-color('japanese-laurel');
+    background-color: get-color(green, light);
+    color: get-color(green);
   }
 
   &--warning {
-    background-color: get-color('peach-cream');
-    color: get-color('brown');
+    background-color: get-color(yellow, light);
+    color: get-color(yellow);
   }
 }
--- a/src/components/MaAlert/MaAlert.scss
+++ b/src/components/MaAlert/MaAlert.scss
@@ -1,6 +1,6 @@
 .alert-banner {
   display: flex;
-  color: get-color(gray, darkest);
+  color: get-color(gray, darker);
 
   &--small {
     padding: rem(4px);
--- a/src/components/MaButton/MaButton.scss
+++ b/src/components/MaButton/MaButton.scss
@@ -15,12 +15,12 @@
     cursor: not-allowed;
 
     &:not(.ma-button--loading) {
-      border-color: get-color('boulder');
-      color: get-color('boulder');
+      border-color: get-color(sky);
+      color: get-color(sky);
 
       &:not(.ma-button--secondary):not(.ma-button--no-background):not(.ma-button--white) {
-        background: get-color('boulder');
-        color: get-color('white');
+        background: get-color(sky);
+        color: get-color(white);
       }
     }
   }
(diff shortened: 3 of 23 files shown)


src/components/MaAlert/MaAlert.scss
```diff
@@ -75,22 +75,22 @@
...
-    color: get-color('japanese-laurel');
+    background-color: get-color(green, light);
+    color: get-color(green);
   }
 
   &--warning {
-    background-color: get-color('peach-cream');
-    color: get-color('brown');
+    background-color: get-color(yellow, light);
+    color: get-color(yellow);
```
afontcu, line 94: Looks like we have some a11y issues here:

image

contrast is 1.37:1, expected 3:1

afontcu: Just noticed that the proposal for Alerts in the Figma file changes them a bit, so that this issue would not exist. Not sure how we should handle the period of time between "big" changes in margarita 🤔

edufarre (approved):
Looks promising! As we discussed we should wait for some feedback from our design team, but overall, looks great! I love the new way we'd handle the colours and their lighter/darker variations.

src/components/MaButton/MaButton.scss
```diff
@@ -43,109 +43,109 @@
...
-      outline: 2px solid get-color('classic-rose');
+      outline: 2px solid get-color(pink, light);
       outline-offset: 0;
     }
 
     &:hover:not(:disabled),
     &:active:not(:disabled) {
-      background-color: darken(get-color('dark-hot-pink'), 10%);
-      color: get-color('white');
+      background-color: darken(get-color(pink), 10%);
```
sergigraciaj, line 65: For me it's a bit confusing using `darken(get-color(pink), 10%)` together with `get-color(pink, dark)`, I suppose it's something we will fix when applying the DS colors

sergigraciaj:
I like the fact that now we know the color we are using but I'm not sure about the dark/light naming, I would use the original naming, for example `get-color(yellow, peach-cream)`. I think this solves the issue of having more than three colors (for example, in the DS with greys).
What about gradients? Could we use a similar approach? Like `get-gradient(pinkToYellow)`, what do you think?

src/components/MaButton/MaButton.scss
afontcu: this change comes from a search&replace. The output remains the same, colors should change once we implement new design system' buttons

afontcu:
> I like the fact that now we know the color we are using but I'm not sure about the dark/light naming, I would use the original naming, for example `get-color(yellow, peach-cream)`. I think this solves the issue of having more than three colors (for example, in the DS with greys).

Hum, not entirely sure how this woudl solve the issue of having more names than needed? You would still need to know "peach-cream" is something.

afontcu:
> I think this solves the issue of having more than three colors (for example, in the DS with greys).

Agree that having "sky" and "ink" might be complicated, even though I feel it follows a logic: "ink" is gray colors related to text (thus "ink"), and sky and lighter grays. tl;dr: `from #000 to #888 --> sky realm, from #888 to #fff -> sky realm`

afontcu:
>  What about gradients? Could we use a similar approach? Like get-gradient(pinkToYellow), what do you think?

Since they are not defined as global variables now, I'd say we should stick to colors here (as they are a big enough breaking change) and handle gradients in a future non-breaking release

afontcu:
> I think this solves the issue of having more than three colors

Also, having "only" three colors is something I just decided. It's not set in stone, as we could set as many variations as needed 🤷‍♀️ it just felt right because there's less naming to learn about, but that's it!

src/components/MaAlert/MaAlert.scss
afontcu: okay, kinda fixed that for the current iteration:

image

obviously, alerts would require some additional love once tokens (colors, spacing…) are defined. For now, at least, our alerts will be a11y compliant.

crisbusquets: Works for me 🚀

edufarre:
> > I think this solves the issue of having more than three colors
> 
> Also, having "only" three colors is something I just decided. It's not set in stone, as we could set as many variations as needed 🤷‍♀️ it just felt right because there's less naming to learn about, but that's it!

Actually I believe this is the way to go. As you all already know, I used to work with Vuetify, and I really liked the way the have it there, it was super fast to catch up and look for the appropriate colour. 

IMHO, I think having just a few 'main colours (red)' and then 'some variations (light)' of them is the best approach 

See how they use it --> https://vuetifyjs.com/en/styles/colors/

src/assets/icons/warning.svg
```diff
@@ -1,3 +1,3 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 22 19" fill="#945d00">
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 22 19" fill="#f5d154">
```
afontcu, line 1: disclaimer: the minute SVGs are handled properly (= not as background images, but as SVGs 🤷‍♀️) will be able to use `currentColor` and ditch the hex value. This is beyond the scope of this PR.

afontcu:
Not sure how to handle the "100% verde" greens. I can see how #1F6D4A could become `green dark`,  but I'm struggling with #24B578 (the "eucaliptus" one).

Ideas? Could we create a new hue map?

edufarre:
> Not sure how to handle the "100% verde" greens. I can see how #1F6D4A could become `green dark`, but I'm struggling with #24B578 (the "eucaliptus" one). It's just a different hue.
> 
> Ideas? Could we create a new hue map?
> 
> image

You're right, but I believe that if we 'isolate' the colours by they 'semantic base name' in order to have a better colours structure, we should include the `eucalyptus` inside the `green` colour, without caring about the `hue`. However, dunno if this is the best approach🤔

Samitier:
I like it! I think the classification is way clearer that way and the implementation makes `get-color` actually useful.

That being said I still vote for the simplest implementation and nomenclature: no `get-color`, just plain scss/vanilla css variables: `red0`, `red1`, `red2`, etc. Just because I still feel like I will have to go search if the color I want is `-er` , `-est`, `light` or `dark` instead of just writing code. Also because I think a unique variable name will be easier to search & replace. But just personal preference.

Other stuff that comes to mind:

 - Now that we are at it, do we want the colors accessible from the JS? For example to not hardcode colors if we need to pass them to some component: `<whatever :color="$colors.blue.light" />`.
 - Should we redesign the storybook colors page so it makes sense in this PR or is it out of scope?
 - Will gradients/box shadows also be scss variables? I vote yes.

Samitier (approved):

afontcu:
> That being said I still vote for the simplest implementation and nomenclature: no get-color, just plain scss/vanilla css variables: red0, red1, red2, etc. Just because I still feel like I will have to go search if the color I want is -er , -est, light or dark instead of just writing code

The goal of `get-color` is to fail hard if you're specifying a made-up color. Having red0, red1, red2 variables won't help us that much there, because not all colors have the same amount of tones (there's only one orange, for instance).

That being said, I don't have a reaaaally strong opinion on this, so happy tweak it xD.

> Now that we are at it, do we want the colors accessible from the JS? For example to not hardcode colors if we need to pass them to some component: `<whatever :color="$colors.blue.light" />`.

Hm, I haven't found the need for it until now, but that's totally doable. I guess we could even add it in future releases if we find the need?

> Should we redesign the storybook colors page so it makes sense in this PR or is it out of scope?

I just "fixed" it in order the display the proper colors, but it is true it could use some rework. I guess this is OOS since we have the Figma file, and because any change in docs won't require a release so we're not in a hurry?


> Will gradients/box shadows also be scss variables? I vote yes.

Yeah, OOS from this PR but it makes total sense to handle them from here, too.

carmenansio (approved):

afontcu:
Added brown hue. Breaking changes from last beta:

`get-color(gray, lighter)` is now `get-color(brown, light)`

| old              | new          |
|------------------|--------------|
| concrete         | brown light    |
| wild-sand        | brown ligth        |

PR #273, opened 2020-09-02
feat: Add MaStack + responsive plugin

### LINKED ISSUE #256: Add vertical layout component for the new Design System
## What

Add a vertical layout component that allows us to stack and space UI elements using the spacing properties of the Design System:

image

## How (let's discuss it)

This discussion has been going on for a while in the [Stack component RFC](https://github.com/holaluz/margarita/pull/251). I think it's better to continue it there.

## Tasks

- [ ] Decide the "how".
- [ ] Implement new approach.
- [ ] Document it.

## Breaking changes

No breaking changes.

### LINKED ISSUE #266: Responsive plugin in margarita?
In `maitai` we had to implement a UI that was pretty different for mobile and desktop:

image

For example, the power section being outside the tabs in mobile, and mobile having a bottom bar with the price.

Since showing/hiding content with mediaqueries would mean adding tons of css and also losing a lot of performance (since we would have entire sections in the dom with display none at all times) we decided to hide/show components with a vue plugin.

The code in the templates ended up being something like this:

```vue
<page>
  <component-only-on-mobile  v-if="$layout.isMobile"/>
  <component-only-on-desktop v-if="$layout.isDesktop" />
  <shared-component />
</page>
```

And the vue plugin `layout.js` being this:

```js
import Vue from 'vue'
import throttle from 'lodash/throttle'

const throttleMs = 100
const mobileBreakpoint = 768

const isMobile = {
  install(vue) {
    vue.prototype.$layout = Vue.observable({
      isMobile: false,
      isDesktop: true,
    })

    if (process.server) return
    checkBreakpoint()
    window.addEventListener('resize', throttle(checkBreakpoint, throttleMs))

    function checkBreakpoint() {
      const isMobile = window.innerWidth < mobileBreakpoint
      vue.prototype.$layout.isMobile = isMobile
      vue.prototype.$layout.isDesktop = !isMobile
    }
  },
}

Vue.use(isMobile)
```

What do you think? Would you like this on margarita? Maybe adding more breakpoints in it like `isTablet` or directly with the nonenclature `xs`, `sm`, `md`, etc?

### LINKED ISSUE #268: [RFC] Add Stack component
Superseeds #251 

Related to https://github.com/holaluz/margarita/issues/256

---

I've explored some simple API for a Stack component that handles spacing in a composable way, and this is what I came up with. Please do only **focus on layout**, not styling!! :)

## An example

Original:

image

Result:

image

(there's just some stylistic CSS, but the important bits is that different elements have different spaces between them – all by composing the same Stack component)

the code for the image above:

```vue
<template>
    <stack space="medium">
      <stack align="center">
        <h1>acceso a ventas</h1>
      </stack>
      <stack space="small">
        <label>email</label>
        <input />
        <label>contraseña</label>
        <input />
        <p>recuperar mi contraseña</p>
        <button>entrar</button>
      </stack>
    </stack>
</template>
```

---

## Another example

Original: 
image

Result:

image

Snippet:

```vue
<template>
  <stack space="medium" align="center">
    <stack space="small">
      <h1>introduce tu dirección CUPS</h1>
      <p>esta información es necesaria porque blablabla</p>
    </stack>

    <!-- Same as the original proposal: This one should have a container that aligns right -->
    <p style="align-self: end;">direccion / cups toggle</p>
    <stack space="small">
      <cups-input />
    </stack>
    <button>descubre tu tarifa</button>
    <p>volver atrás</p>
  </stack>
</template>
```

Notice how the spacing is different between the title and the subtitle, and the rest of elements from the form.

## Proof of Concept implementation

```vue
<template>
  <component :is="component" :style="style" class="stack">
    <slot />
  </component>
</template>

<script>
export default {
  name: 'Stack',

  props: {
    space: {
      type: String,
      required: true,
      validator: (p) => ['small', 'medium', 'large'].includes(p),
    },

    align: {
      type: String,
      default: '',
      validator: (p) => !p || ['start', 'center', 'end'].includes(p),
    },

    component: {
      type: String,
      default: 'div', // this way we could handle lists and so
    },
  },

  computed: {
    // to be replaced with a proper implementation of DS tokens
    spaceValue() {
      if (this.space === 'small') return '0.5rem'
      if (this.space === 'medium') return '2rem'
      return '3rem'
    },

    style() {
      return {
        '--space': this.spaceValue,
        '--align-items': this.align,
      }
    },
  },
}
</script>

<style scoped>
.stack {
  display: flex;
  flex-direction: column;
  align-items: var(--align-items);
  gap: var(--space);
}
</style>
```

Remember that this is just a prototype to see how well it handles common scenarios 😉 

---

Thoughts? Would this solve the scenarios you face on a daily basis? What would you miss from it?

afontcu:
This PR closes #268, closes #266, closes #256, and closes #251 🎉 

This PR implements two complementary things which, unfortunately enough, were quite hard to implement & test independently. Sorry about that. In return, I will happily walk you through decisions I made, feel free to ask.

---

# Responsive plugin

We've leveraged [MatchMedia](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia) (lovely) and a Vue plugin because it will make it easy to migrate in a future Vue 3 bump – we will just need to move it to a composition function, not that tricky.

The plugin exposes `this.$layout.currentBreakpoint` to our margarita-consuming apps, so we can use their value to tweak UIs accordingly.

# MaStack

The long-awaited(?) Stack component, which handles **vertical layouts**. It has been largely discussed previously (#268, #256, #251). You might want to check out locally to this branch and play with storybook.

Also, it comes with responsive props in conjunction with the Responsive plugin:

stack

I used CSS Grid because there's still little support for `gap` in flexbox. That's a pity, but it is also the world we live in xD

Also, I created a `tokens.js` file because I felt some of these stuff should live in a somewhat "global" place. Feel free to suggest other structures :) 

---

## Out of scope

- Handling horizontal spacing (another component should do that – MaColumn?)
- Properly handle tokens (they are painfully split between JS and SCSS, for now)

## To be defined
- ~Re https://github.com/holaluz/margarita/issues/268#issuecomment-684747024, how should be handle full-width situations? is something Stack should take into account? What's the desired default behavior? Ideas?~ update: done https://github.com/holaluz/margarita/pull/273#issuecomment-687074702
- ~Naming alignment props "start,center,end" makes sense? Should we map it to "left, center, right"?~ update: yes


src/utils/responsivePlugin.js
```diff
@@ -0,0 +1,63 @@
...
+function getCurrentBreakpoint(tabletQuery, desktopQuery) {
+  if (desktopQuery.matches) {
+    return breakpointsEnum.desktop
+  }
+
+  if (tabletQuery.matches) {
+    return breakpointsEnum.tablet
+  }
+
+  return breakpointsEnum.mobile
```
afontcu, line 62: mobile by default

src/scss/variables/_spacing.scss
```diff
@@ -0,0 +1,10 @@
+$spaces: (
+  none: 0,
+  xxsmall: 4px,
+  xsmall: 8px,
+  small: 12px,
+  medium: 16px,
+  large: 24px,
+  xlarge: 32px,
+  xxlarge: 48px,
```
afontcu, line 9: totally made-up. it should match Figma's `layout spacing` tokens.

src/components/MaStack/MaStack.vue
```diff
@@ -0,0 +1,53 @@
...
+    },
+
+    computedAlign() {
+      return this.$layout.getResponsivePropValue(this.align)
+    },
+
+    classes() {
+      return {
+        [`stack--space-${this.computedSpace}`]: true,
+        [`stack--align-${this.computedAlign}`]: !!this.align,
```
Samitier, line 48: `align` has a default, no? Will it ever be falsy?

sergigraciaj (commented):
Wow, such a great job!

src/utils/responsivePlugin.js
```diff
@@ -0,0 +1,63 @@
+import Vue from 'vue'
```
Samitier, line 1: if this is a plugin why is it in `/utils` and not in `/plugins`?

src/components/MaStack/MaStack.scss
```diff
@@ -0,0 +1,19 @@
+$stack-spacing: map-keys($spaces);
+$stack-alignment: (start, center, end);
+
+.stack {
+  display: grid;
```
sergigraciaj, line 5: In the RFC it was using flex, why the change to grid?

src/utils/responsivePlugin.js
```diff
@@ -0,0 +1,63 @@
...
+import { breakpoint, breakpointsEnum } from '../tokens'
+
+export default {
+  install(vue) {
+    vue.prototype.$layout = Vue.observable({ currentBreakpoint: null })
+    vue.prototype.$layout.getResponsivePropValue = getResponsivePropValue
+
+    if (process.server) return
```
Samitier, line 9: what will happen with SSR if `currentBreakpoint` is `null`? Shouldn't we leave one breakpoint by default just to render something in the server?

src/components/MaStack/MaStack.vue
```diff
@@ -0,0 +1,53 @@
...
+    align: {
+      type: [Array, String],
+      default: alignment.start,
+      validator: responsivePropValidator(alignment),
+    },
+
+    component: {
+      type: String,
+      default: 'div',
+      validator: (p) => ['div', 'ol', 'ul'].includes(p),
```
sergigraciaj, line 32: We won't be using semantic tags here? It's different the use of a composable component vs a semantic tag?

src/utils/responsivePlugin.js
```diff
@@ -0,0 +1,63 @@
+import Vue from 'vue'
+import { breakpoint, breakpointsEnum } from '../tokens'
+
+export default {
+  install(vue) {
+    vue.prototype.$layout = Vue.observable({ currentBreakpoint: null })
+    vue.prototype.$layout.getResponsivePropValue = getResponsivePropValue
```
Samitier, line 7: will we use `currentBreakpoint` or only `getResponsivePropValue`? Maybe one of those is not necessary?

src/tokens.js
```diff
@@ -0,0 +1,23 @@
+export const breakpointsEnum = {
```
Samitier, line 1: maybe this file is too tied with the responsive plugin and they could both live in the same directory? Like the """""types"""""" file of the responsive plugin 😂

src/utils/responsivePropValidator.js
```diff
@@ -0,0 +1,7 @@
+export function responsivePropValidator(validValues) {
```
Samitier, line 1: Maybe we can create some kind of custom type for the prop validator, like:

```js
props: {
    align: {
        type: ResponsiveProp
   }
}
```

I don't know if its possible or if it solves anything in our use case. I only found [this](https://vuejs.org/v2/guide/components-props.html#Type-Checks). 

Maybe we could use the `normalizeResponsiveProp` in that type constructor too so the components only receive the array of breakpoints regardless of the input.

Samitier:
>Re #268 (comment), how should be handle full-width situations? is something Stack should take into account? What's the desired default behavior? Ideas
Naming alignment props "start,center,end" makes sense? Should we map it to "left, center, right"? 

To solve both problems: I vote "left, center, right, fill", where fill is full-width for all children. I would completely ignore the text-align prop and let every children text-align its content by themselves.

Samitier (commented):
Really nice work, btw! 😃

src/components/MaStack/MaStack.scss
afontcu: from the PR description: "_I used CSS Grid because there's still little support for gap in flexbox. That's a pity, but it is also the world we live in xD_"

src/tokens.js
afontcu: you mean the breakpoints enum or all the breakpoint-related stuff? The enum, yeah, you're right, they could live _inside_ the plugin file, to be honest. We will have two `'tablet'` strings in the codebase, but that's far from critical xD Should I make the change, then?

src/utils/responsivePropValidator.js
afontcu: Ah, that'd be great 🤔 we'd allow string or arrays, but then we can always handle it as a `ResponsiveProp`. I'll take a look and see if I can come up with a solution – yet, since this won't mean any external change, we can always iterate on it ;)

src/components/MaStack/MaStack.vue
afontcu: yeah, actually this is kind of half-baked… my idea was that, if OL or UL were provided, Stack should wrap its children in LI tags.

```js
<ma-stack component="ul">
  <something />
  <something-else />
</ma-stack>
```

to

```html
<ul>
  <li><!-- whatever "something" renders --></li>
  <li><!-- whatever "something-else" renders --></li>
</ul>
```

this is why specifying OL or UL made sense.


We can ditch this for the time being (remove the `component` prop altogether, actually) and see if that's a real need.

src/components/MaStack/MaStack.vue
afontcu: yep, changed the default value and forgot to made the appropriate change here ^^ nice catch

src/utils/responsivePlugin.js
afontcu: good point. Gonna move it

src/utils/responsivePlugin.js
afontcu: good point – what's happening right now with ur current implementation in maitai? Should we mimic that behavior?

src/utils/responsivePlugin.js
afontcu: Hm –

`currentBreakpoint` provides you with medium|tablet|desktop string, so you can use it in your app's logic.
`getResponsivePropValue` provides you with a function that, given an array of potential prop values, gives you the appropiate one depending on `currentBreakpoint`.

One depends on the other one, but they serve different purposes: `getResponsivePropValue` will be heavily used in all Layout components (and potentially in other places, too), while `currentBreakpoint` will shine in other apps (not in margarita, but its consumers).

src/components/MaStack/MaStack.vue
afontcu: I think I'll ditch `component` prop, so we can discuss this in future iterations. For now, I guess

```vue
<ma-stack>
  <ul>
    <li></li>
    <li></li>
  </ul>
</ma-stack>
```

would do the trick. We can't nest DIVs inside ULs because that's invalid markup, so the wrapping becomes a bit weird. That was my first idea, which then evolved to "okay, so now let's wrap children with LI elements".

src/utils/responsivePlugin.js
Samitier: in maitai, SSR is always considered as desktop. The hypothesis was that desktop pages always tend to have more text in them and they are better for SEO, but I dunno if it makes a difference 😂

src/utils/responsivePlugin.js
Samitier: I thought the idea was to "hide" the plugin. But yeah, I also prefer to be able to use it outside margarita, so nevermind!

src/utils/responsivePlugin.js
afontcu: hahahaha lol, okay. let's set `vue.prototype.$layout.currentBreakpoint` to `'desktop'` by if we're in server

src/tokens.js
Samitier: I was refering to everything 😂 

And maybe make the "public" stuff visible through the plugin itself (i.e: `this.$layout.breakpoints`) so we can check them from outside margarita too?

afontcu: hm, for instance, spacing and alignment tokens are kind of global, right? Several other components may want to use it. Where would you think it'd make sense to store them?

src/utils/responsivePlugin.js
afontcu: it would be great to toooootally hide the plugin, but I fear that sooner than later we'll need some custom tweaks – the design system + responsive props won't be able to handle everything!

src/components/MaStack/MaStack.scss
afontcu: https://caniuse.com/#feat=flexbox-gap

src/tokens.js
Samitier: inside the plugin maybe? And only access them using this.$layout instead of import. But I'm kinda just guessing at this point, I think we better wait to see what we really need.

Samitier (approved):

src/components/MaStack/MaStack.vue
afontcu: I went ahead and remove `component` prop altogether – we can always bring it back as an additive PR

afontcu:
> > Re #268 (comment), how should be handle full-width situations? is something Stack should take into account? What's the desired default behavior? Ideas
> > Naming alignment props "start,center,end" makes sense? Should we map it to "left, center, right"?
> 
> To solve both problems: I vote "left, center, right, fill", where fill is full-width for all children. I would completely ignore the text-align prop and let every children text-align its content by themselves.

What about this:

* Stack always sets `start` (should we call it `left`????) by default.
* This means that fluid UI components (for instance, a button with width:100% by default) will fill the whole available space.
* Also, width-constrained components will fill the specified space. No need to handle it from Stack.
* Some components (the ones we find the need to) will feature an `align` prop, so that we can handle their alignment the same way we handle Stack's:

```vue
<ma-stack>
  <ma-button>Click me</ma-button> <!-- aligned left because of Stack, but since it is fluid, it fills the whole thing -->
  <ma-font>A random text</ma-font> <!-- text-align:left by default, as usual -->
  <ma-font :align="['middle', 'left']">foo</ma-font> <!-- text-align:center on mobile, left above it -->
</ma-stack>
```

OTOH, Stack's align prop always trumps child's:

```vue
<ma-stack align="right">
  <ma-font>A random text</ma-font> <!-- right aligned -->
</ma-stack>
```

src/index.js
```diff
@@ -32,12 +34,14 @@ export {
...
   MaSelect,
   MaSidebar,
+  MaStack,
   MaText,
   markdown,
   markdownSSR,
 }
 
 function install(Vue) {
+  Vue.use(responsivePlugin)
```
afontcu, line 44: Is there a better way to provide the `responsivePlugin` to margarita-consuming apps?

afontcu: @Samitier what do you think of this approach for projects such as kalimotxo or any non-vue app such as the blog or unbounce? 

related to https://github.com/holaluz/kalimotxo/issues/27, I guess!

carmenansio (approved):

sergigraciaj (approved):

PR #277, opened 2020-09-08
Bump deps (Including Storybook v6!)

afontcu:
Latest versions of Storybook come with [several changes](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md), the biggest one being a new format to express stories (called CSF – [more on that](https://medium.com/storybookjs/component-story-format-66f4c32366df)).

I've ran a couple of codemods to update our current stories, but they should be revisited once we update each component to match the Design System. Button is the one I tweaked a bit more, because it was kinda messy (it still is, to be honest 😂).

CSF comes with several benefits which I haven't explored here. For instance, we could share the same component instances between stories and tests ([source](https://storybook.js.org/docs/react/workflows/unit-testing)), which is amazing.

Also, I believe [Controls](https://storybook.js.org/docs/vue/essentials/controls) would make stories way more powerful/cool instead of relying on manual knobs. ~I'll leave this PR as a draft, and explore this path a bit.~ edit: too much to add in a single PR

Finally, now it is really easy to create sections :D

Captura de Pantalla 2020-09-08 a les 13 55 44

---

Other bumpings yielded nothing worth mentioning 👍


Samitier (approved):
The ``

afontcu:
> The major of `snarkdown` doesn't break the script/hack we did in `v-markdown`, right? Cuz it was pretty hacky hahahh

Tests passed, so… 😬

afontcu:
> The major of `snarkdown` doesn't break the script/hack we did in `v-markdown`, right? Cuz it was pretty hacky hahahh

Tests are passing, so I'll go ahead and merge this PR as a non-release! If anything breaks related to v-markdown in a near future, it's on me XD

PR #279, opened 2020-09-23
Fix: fix color CSS output

afontcu:
This PR fixes an issue with the generated CSS (see the following [thread in slack](https://holaluz.slack.com/archives/C019DLZTBU6/p1600701743003300?thread_ts=1600675842.001100&cid=C019DLZTBU6)).

tl;dr: some sass loader was transforming `white-base` to `#fff-base`, thus breaking the `:export` statement. Since that statement was only meant to provide colors to the storybook story, I removed it from the CSS output and put it back inside storybook.

Also I took the liberty to improve the colors page _a bit_:

image

### DIFF
--- a/src/components/colors.stories.js
+++ b/src/components/colors.stories.js
@@ -25,12 +25,14 @@ export const Colors = () => ({
         <h2 :style="titleStyle">{{hue}}</h2>
         <div v-for="(hex, tone) in tones" :key="tone" :style="colorStyle">
           <div :style="getBoxStyle(hex)">
-            <p :style="colorNameStyle">{{hue}} {{tone}}</p>
-            <code v-text="hex" />
             <code v-text="getFunction(hue, tone)" />
           </div>
+          <p :style="colorNameStyle">
+            {{hue}} {{tone}}
+            <code v-text="hex" :style="hexStyle" />
+          </p>
         </div>
-        </div>
+      </div>
     </div>
   `,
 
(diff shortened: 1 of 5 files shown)


src/scss/variables/_colors.scss
```diff
@@ -4,57 +4,49 @@ $colors: (
...
 $shadow-medium: rgba(0, 0, 0, 0.0980392);
 $shadow-dark: rgba(0, 0, 0, 0.498);
-
-:export {
-  @each $name, $colorGroups in $colors {
-    @each $tone, $value in $colorGroups {
-      #{$name}-#{$tone}: $value;
-    }
-  }
-}
```
afontcu, line 60: These are the line changes that fix the build. @Samitier is there a way to test it out with kalimotxo?

src/components/colors.stories.js
```diff
@@ -1,4 +1,18 @@
-import colors from '../scss/_margarita.scss'
+import merge from 'lodash.merge'
+import '../scss/_margarita.scss'
```
afontcu, line 2: btw; this line loads Margarita for all stories. There's no easy way to load global SCSS in Storybook (and tbh I didn't want to spend much time on this). Feel free to pick it up and try to fix it :D

Samitier: why do we need `margarita.scss`? 

If it's for the color variables, can't `colors.scss` import the color variables directly? I mean doing the `@import 'variables/colors` from inside the scss.

src/scss/variables/_colors.scss
Samitier: I can build margarita locally and overwrite the node_modules dist folder in kalimotxo. Gimme a sec.

src/components/colors.stories.js
afontcu: Otherwise, storybook loads Margarita components, but without the CSS. For instance, fonts and normalization was missing 😞  as far as I can tell, this is similar to https://github.com/holaluz/bonasera/blob/master/src/ui/nuxt.config.js#L27

src/scss/variables/_colors.scss
Samitier: works perfectly 🎉

Samitier (approved):
🔝🔝🔝🔝 thank you!

src/components/colors.stories.js
Samitier: you could import a global css with the theming api, but I don't think you can import a scss. Meh, I wouldn't lose that much time with this.

afontcu: > but I don't think you can import a scss

yeah, tried to do that by prepending the loaders within the import statement, extending the webpack config… without success. So I just import it manually and called it a day 🤷

src/scss/variables/_colors.scss
afontcu: great! Imma release the patch version then

PR #291, opened 2020-11-25
chore: fix jest .vue imports & "too many files" issue

Samitier:
Fixes some problems that appeared in https://github.com/holaluz/kalimotxo, which shares the exact same `jest.config.js` as margarita.

More info: https://holaluz.slack.com/archives/CR9670R5H/p1606311768011300

Bonus: fix lint errors in `MaButton` test

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,5 @@
 module.exports = {
-  moduleFileExtensions: ['js', 'jsx', 'json', 'vue'],
+  moduleFileExtensions: ['vue', 'js', 'jsx', 'json'],
   transform: {
     '^.+\\.vue$': 'vue-jest',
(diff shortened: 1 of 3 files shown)


edufarre (approved):
Quick question, wouldn't be better to use `findBy` instead of `getBy`?

afontcu (approved):
(I wouldn't tag this as a `fix:` – and thus an npm release– because it has no effect on the deployed code. Instead I'd tag it as a "chore: jest .vue imports…" which won't trigger an npm release 😉)

afontcu:
> Quick question, wouldn't be better to use `findBy` instead of `getBy`?

hm why? It is a sync query, not an async one

PR #300, opened 2020-12-14
feat: Add types for Margarita plugin

afontcu:
Note: **this PR does not introduce TypeScript to margarita** (yet… *evil laugh*). It simply provides types for basic plugin installation (that is, being able to call `vue.use(Margarita)` in a TS project) and also types for responsive plugin when running on a TypeScript project:

image

🎉🎉 

I've also exposed `Breakpoint` because it might come in handy:

imatge

---

Also, I've added a basic `jsconfig.json` file so VSC can properly understand aliases within Margarita itself. This works even without TS:

imatge


afontcu:
Implementation: https://github.com/holaluz/nuxt-clean/pull/26

types/margarita.d.ts
```diff
@@ -0,0 +1,6 @@
+export declare type Breakpoint = 'mobile' | 'tablet' | 'desktop'
+
+export type Layout = {
+  readonly currentBreakpoint: Breakpoint | null
```
afontcu, line 4: tbh I'm not sure if this could be `null` 🤷  it is initialized as null (https://github.com/holaluz/margarita/blob/master/src/plugins/responsivePlugin.js#L5), but it gets an initial value right after (https://github.com/holaluz/margarita/blob/master/src/plugins/responsivePlugin.js#L21)

sergigraciaj (approved):

PR #302, opened 2020-12-16
chore: Use generic color and spacing design tokens

### LINKED ISSUE #301: Shared design tokens between S/CSS and JS
- Implementation PR: #302 (Just in case you wanna play around)

# Summary

This RFC aims to solve an issue I bumped into lately because of having the **DS color tokens** on a `.scss` file

When developing the `MaText` component, I bumped into kinda important issue. I was intending to create a component that would handle its own content `color` by using the `component props`, just allowing to set the **already defined** colors on the Design System.
The problem here is that our color tokens only existed on a `.scss` file, and I did not know how to manage them in order to use them as `property validators` on the new desired component API, because I wasn't able to check them, find below an example:
```js
...
const DS_COLORS = ?????
...
// Component props
props: {
  color: {
    type: String,
    default: 'gray'
    validator: (color) => DS_COLORS.includes(color)
  }
}
...
```

# Motivation

As I already said, this new approach will help us sharing these tokens across our `.scss` and `.js` files, thus having **a single source of truth** in the entire project. 

This approach would use [Style Dictionary](https://amzn.github.io/style-dictionary/#/) for computing the desired styling files (be aware that we can set the output files as `.scss` and `.css` --> That means that [if we want to get rid of the scss](#298 ) it would practically have no cost when talking about handling our design tokens). Find below a StyleDictionary config and output example:

### SCSS
Config
```js
...
 {
    destination: '_colors.scss',
    format: 'scss/map-flat',
    mapName: 'colors',
    filter: { type: 'color' },
 },
 ...
```
Output
```scss
$colors: (
  'pink-light': #f8b0cb,
  'pink-base': #e6007d,
);
```
### CSS
Config
```js
...
 {
    destination: '_colors.scss',
    format: 'css/variables',
    mapName: 'colors',
    filter: { type: 'color' },
 },
 ...
```
Output
```css
:root {
  --pink-light: #f8b0cb;
  --pink-base: #e6007d;
}
```

The tokens would have to be defined in a `.json` file (find below an example). You could ignore the `type: 'color'` attribute, as it just for the library that helps us isolate the different tokens by their type when creating the style files.
```json
//colors.json
{
  "pink": {
    "light": {"value": "#f8b0cb", "type": "color"},
    "base": {"value":"#e6007d", "type": "color"},
  },
}
```
```json
//spacing.json
{
  "none":{"value": "0", "type": "spacing"},
  "xxsmall":{"value": "1rem", "type": "spacing"},
   ...
}
```

In addition, we could seize our current `npm start` command and implement a watcher of our design token files, thus building the styles each time they change and having the styles always updated. This would use a library named [Chokidar CLI](https://www.npmjs.com/package/chokidar-cli).

```json
...
  "scripts": {
     ...
    "start": "npm run storybook:serve & npm run watch:tokens",
    "build:tokens": "style-dictionary build --platform scss --config ./styleDictionary.config.js && prettier 'src/scssTokens/*.scss' --write",
    "watch:tokens": "npm run build:tokens && chokidar  \"tokens/*.json\" -c \"npm run build:tokens\""
  },
...
```

💡  I believe we should add the `build:tokens` in our margarita's  `"pre-commit": "lint-staged"` hook in order to have them updated when committing changes to the project.

# Adoption strategy

We would have to adapt/change some `scss/functions` in order to handle the newly created tokens architecture, but it would be compatible with the current implementation in our projects. Which means that we could still use our `get-color()` function for example.

The **Style Dictionary** library could also help us when managing our assets – like images, icons and fonts – directly as design tokens. [Example repo](https://github.com/amzn/style-dictionary/tree/master/examples/advanced/assets-base64-embed)

edufarre:
Closes #301 

If you run `npm start` the token build will happen automatically with hot-reload included.
If you just want to check how it works, run `npm run watch:tokens`


edufarre:
Beware of the `.gitignore` file. We currently added the new auto-generated token paths with the intention of moving all of them outside the `/src` folder on a future PR.

```
# auto generated scss token files
src/scss/variables/_colors.scss
src/scss/variables/_spacing.scss
```

src/scss/variables/_shadows.scss
```diff
@@ -0,0 +1,4 @@
+// Temporary. How do we handle shadows?
+$shadow-light: rgba(0, 0, 0, 0.027451);
+$shadow-medium: rgba(0, 0, 0, 0.0980392);
+$shadow-dark: rgba(0, 0, 0, 0.498);
```
Samitier, line 4: hmm this are not actually shadows, just blacks with opacity. Maybe we should just unify the entire box shadow in a variable or treat the shadow color as just another color.

I mean this as box shadow:
```
$shadow-light: npx npx npx rgba(0, 0, 0, 0.027451);
```

Or this as color:
```
shadow-base: rgba(0, 0, 0, 0.027451);
shadow-light: ...
shadow-dark: ...
```

Samitier (approved):
Just out of curiosity, looking at the github actions this feature doesn't look like it slows build times at all, so great job!

src/scss/variables/_shadows.scss
edufarre: I just copied/pasted the variables we already had. I believe this is out of the scope of this PR. However, we should discuss this with our Design peers in order to iterate it.

Samitier: hahah yes, yes. I was just answering the comment in the code 😂

package.json
```diff
@@ -15,16 +15,19 @@
...
     "lint:js": "eslint 'src/**/*.{js,vue}'",
-    "start": "npm run storybook:serve",
+    "start": "npm run storybook:serve & npm run watch:tokens",
     "storybook:build": "vue-cli-service storybook:build -c .storybook",
     "storybook:serve": "vue-cli-service storybook:serve -p 6006 -c .storybook",
     "test": "vue-cli-service test:unit",
-    "test:watch": "vue-cli-service test:unit --watch"
+    "test:watch": "vue-cli-service test:unit --watch",
+    "build:tokens": "style-dictionary build --platform scss --config ./styleDictionary.config.js && prettier 'src/scss/variables/*.scss' --write",
+    "watch:tokens": "npm run build:tokens && chokidar  \"src/tokens/*.js\" -c \"npm run build:tokens\""
```
afontcu, line 30: A small nitpick: shouldn't these be written as `tokens:build` and `tokens:watch` to match other scripts? (`test:watch` or `storybook:build`)

edufarre: You're damn right! My bad, gonna change it!

afontcu: Also: should tokens be generated after installing (say, the first time you download margarita, for instance)? We could do so by adding `postinstall: npm run tokens:build`

edufarre: well, I blieve this is replace by adding the `npm run tokens:build` on the `npm start` command, as the first time you will execute the app, they will be generated :)

afontcu: ah ok, makes sense 👌 didn't notice that watch also triggers build (which is a bit weird if I may, but not that critical)

src/scss/functions/_colors.scss
```diff
@@ -5,19 +5,20 @@
 /// @return {Color} The color value.
 
 @function get-color($hue, $value: base) {
-  $colorMap: map-get($colors, $hue);
+  // Inspecting $name prevents node-sass from
```
afontcu, line 8: 🙈 

```suggestion
  // Inspecting $hue prevents node-sass from
```

afontcu: (what an epic bug we fixed here, by the way. my god 😂)

sergigraciaj (approved):

afontcu (approved):

src/scss/functions/_colors.scss
edufarre: Thanks god you got my back here!

G.. I would have gone insane trying to solve this 🤣

src/scss/functions/_colors.scss
```diff
@@ -5,19 +5,20 @@
...
-  @if type-of($fetched-color) == color {
-    @return $fetched-color;
+  @if type-of($colorMap) == color {
+    @return $colorMap;
   } @else {
-    @error 'Color `#{$hue} - #{$value}` not found. Available colors: #{available-names($colors)}';
+    @error 'Color `#{$hue}-#{$value}` not found. Available colors: #{available-names($colors)}';
   }
+
+  @return colorMap;
```
afontcu, line 21: is this expected? Looks like a leftover

edufarre: It actually is a leftover :shame:

Sorry

afontcu: no worries! fixed

afontcu:
Can we merge this one? 🚀  I'd like to explore the path that this unveils when it comes to remove SCSS 😬

edufarre:
Yes, we can! I tried to do it yesterday but I bumped into other tasks and forgot about this. My bad

PR #303, opened 2020-12-16
feat: Add MaModal

### LINKED ISSUE #297: [RFC] MaModal component
- Figma link: https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/Design-System-Beta?node-id=2566%3A0
- Release: minor

# Summary

A modal is a component with three main blocks:

* A content block, which features the "real" content of the modal. It is hidden by default, and rendered next to `body` when needed.
* A trigger block, which allows users to display the content block.
* An overlay block, which surrounds the content block. A click on this area will close the content.

It should also feature all a11y goodness described [in the spec](https://www.w3.org/TR/wai-aria-practices-1.2/#dialog_modal) (disclaimer. this docs are _informative_, not mandatory. But still).

# Basic example

```vue
<ma-modal title="Enviar la oferta por e-mail" width="small" header-type="gradient">
  <template #trigger="{openModal}">
    <button @click="openModal">Open the modal</button>
  </template>

  <!-- this will get rendered as a child of <body> -->
  <template #content="{closeModal}">
    <ma-stack>
      <p>My amazing content</p>
      <button @click="closeModal">close</button>
    </ma-stack>
  </template>
</ma-modal>
```

image

You could also use the slot to render a specific component:

```vue
<ma-modal title="My amazing modal">
  <template #trigger="{openModal}">
    <button @click="openModal">Open the modal</button>
  </template>

  <my-amazing-modal-content #content />
</ma-modal>
```

# Motivation

Modals are a common UI pattern that are hard to get right. By providing a base component from Margarita, we make sure we'll have a component that is compliant with out standards of usability, a11y, and maintainability.

# Detailed design

## Props

```js
title: { type: String, required: true }

width: { type: String, default: 'medium', validValues: ['small', 'medium', 'large'] }

headerType: { type: String, default: 'white', validValues: ['white', 'gradient'] }
// to render a gradient heading
```

## Slots

```
#trigger (openModal: Function)
#content (closeModal: Function)
```

## Emits

```
'open'
'close'
```

The `close` emit will be triggered when:
1) the `closeModal` function is executed
2) user clicks on the overlay
3) user clicks escape

## DOM output

The component will take everything from the `#content` slot and move it next to the end tag of `body`. This way, we avoid potential issues when it comes to absolute/relative positioning, and z-indexes. The implementation of this feature can be done by hand, or using [portal-vue](https://github.com/LinusBorg/portal-vue).

If MaModal contains a component, it will remain a logical child component of the MaModal's parent. 

~This also means that injections from a parent component work as expected, and that the child component will be nested below the parent component in the Vue Devtools, instead of being placed where the actual content moved to.~
update: I don't think this is doable in Vue 2 without a lot of magic. This means that if the slot content of MaModal contains components, they will show up as children of MaModal in the devtools.

(Simplified) resulting DOM structure from the previous example:

```html
<html>
  <body>
    <button>Open the modal</button>

    <!-- modal structure is totally made up. Simply notice where it gets rendered -->
    <div class="ma-modal">
      <p class="ma-modal__heading">Enviar la oferta por e-mail</p>
      <div class="ma-stack">
        <p>My amazing content</p>
        <button>close</button>
      </div>
    </div>
  </body>
</html>
```

## Styling
MaModal aims to provide little-to-no styling. It will render the overlay, center the modal, pad its inner content and manage its size, but nothing beyond this point. MaModal does not aim to lay out its inner content other than heading, so consumer should be responsible for it.

In terms of width, `width` prop will set a max width for the modal, and shrink if needed.

In terms of height, the component will grow to wrap its content, and overflow (with scroll) if content grows higher than available vertical space.

MaModal heading will render the provided title value and also a close button which will close the modal. The heading will remain visible even when vertical overflow is present.

## Accessibility

As mentioned, it should feature all a11y goodness described [in the spec](https://www.w3.org/TR/wai-aria-practices-1.2/#dialog_modal). In particular, autofocus on open, focus trap, and appropriate ARIA roles must be there.

# Adoption strategy

While this RFC does not introduce a breaking change, adopting MaModal in projects might require several modifications (depending on current usage). That being said, I cannot foresee any current scenario where MaModal isn't able to match current features of our modals.

# Unresolved questions

- ~Should we allow for multiple MaModal elements to be opened at the same time?~ update: No
- ~Should the `size` prop accept a ResponsiveProp value such as `['small', 'large']`?~ update: No
- ~Are there situations where we need to know if the modal is open, for instance in the `#trigger` slot? Should this info be provided to the slot?~ update: Not for now.

afontcu:
This PR closes #297. This PR aims to serve as an implementation of the RFC: Almost everything from there is built into it.

Some notes:

- I've used [vue-simple-portal](https://github.com/LinusBorg/vue-simple-portal) because is the most simple way to end up with a [Vue3-compatible API](https://github.com/LinusBorg/vue-simple-portal/issues/56).
- The component features several a11y goodies which makes the big picture a big scary. Feel free to reach me (or better, drop a comment!) for additional clarifications.
- I tried to [write tests](https://github.com/holaluz/margarita/pull/303/files#diff-5007d823e5d98623f02c3fd45f67c9254171406ae3b2c52e9a2dae87add935ee) that would serve as documentation. Please check them out to make sure you fully grasp the component!
- It successfully renders the modal content as a direct child of its parent component while rendering the DOM node as a direct child of `<body>` 🎉 

Missing bits:

- [x] Make sure transitions are ok. I guess we'd want two different transitions (one for the overlay and one for the modal itself).
- [x] Make sure intra-spacing is ok.
- [x] Make sure styles are ok in general (I just added some for demonstration purposes).
- [x] Add another story, if needed.
- [ ] It'd be great if we could test that we're not messing with SSR (as [we did there](https://github.com/holaluz/margarita/blob/master/src/plugins/responsivePlugin.node.spec.js)).


Samitier:
I find this code a bit hard to understand 😓 

I thought the benefit of using portals was because it was a feature planned for vue core, but if we have to do it by hand I kinda don't see what we get from it...

Are you sure we want to discard the plugin approach? It works fine and the component has a way simpler API. It also has a trivial implementation.

IMHO the only problem with the plugin approach is that we have to manually instantiate `<fullscreen-modal />` at the root of the project every time, but that is not necessary true. We can (probably) instantiate it programatically from inside the plugin and attach it to `this.$root` the first time we run `$modal.open`. Maybe we should explore that option? Also, the migration path will be easier because I think every project is already using a plugin/store to open modals.

afontcu:
> I find this code a bit hard to understand 😓
> 
> I thought the benefit of using portals was because it was a feature planned for vue core, but if we have to do it by hand I kinda don't see what we get from it...

What parts of the code do you find hard to understand? Is it the internal components (that are copypasted from vue-simple-portal), or all the focus management? The former could be fixed if I could properly import the component from the library, but as I mention I ran into some issues 😞 The latter, well, I admit is complicated to follow, but I'm not sure how to make it simpler. There's a bit of logic there, maybe extracting it to a helper function?


> It works fine and the component has a way simpler API

The API is quite simple, isn't it? 🤔 I mean, it's 3 props and a couple of slots.

Samitier:
> What parts of the code do you find hard to understand? Is it the internal components (that are copypasted from vue-simple-portal), or all the focus management?

Both 😓  . They even introduce concepts I didn't even know existed (abstract components 🤯, `this.$once('hook:destroyed'` ).

And as for the focus trapping, IMHO is way simpler to manage with this approach (unless I'm forgetting edge cases).

When we open a modal:
 - We `focus()` on the close button.
 - We add this listener:
```js
document.addEventListener("focus", ({target}) => {
  if (this.isModalOpen && !this.$refs.modal.contains(target)) {
      event.stopPropagation()
      this.$refs.closeButton.focus()
  }
}, true)
```
In order to "close" the modal, we remove the listener and `.focus()` to the element that opened the modal (we can save its ref when opening).

Also, we should add the accessibility stuff: `tabindex=-1` and `role='dialog'` to the modal.

Here is way better explained: https://www.w3.org/WAI/GL/wiki/Using_ARIA_role%3Ddialog_to_implement_a_modal_dialog_box

> The API is quite simple, isn't it? 🤔 I mean, it's 3 props and a couple of slots.

I mean complex in the sense that it introduces the concept of slot scopes, which we don't normally use (compared to plugins). Also I think I said it in the RFC but the DOM kinda gets a bit weird/hard to read at first glance when the button that opens the modal is inside the modal component itself. 

Also what happens if we want a link in a text to open the modal. Do we have to set the whole text inside the `#trigger`? What if we want to open it programatically through js (IE to show an error)?

afontcu:
> They even introduce concepts I didn't even know existed (abstract components 🤯)

Yeah, I didn't know about abstract components either 😂  As I mentioned, this comes [directly from vue-simple-portal](https://github.com/LinusBorg/vue-simple-portal/blob/develop/src/components/TargetContainer.js#L3-L7), so it is something we shouldn't worry about. I'll try to make it work with installing the plugin so we hide all this complexity in node_modules, thou! Would that help?

> this.$once('hook:destroyed')

Yeah, this is an uncommon feature, I grant you that. It came in handy to handle all the event listening stuff together. It can easily be removed, though, happy to commit the following changes if you feel this would make the code easier to understand/read 👍 :

image




> Also, we should add the accessibility stuff: tabindex=-1 and role='dialog' to the modal.

The role is already being set: https://github.com/holaluz/margarita/pull/303/files#diff-9acc1067e3f78f2f954ef4d55504b7002f0130bbb2375e3784489e2e70006af3R24
and tested: https://github.com/holaluz/margarita/pull/303/files#diff-5007d823e5d98623f02c3fd45f67c9254171406ae3b2c52e9a2dae87add935eeR210

re tabindex=-1 and focusing on close button, I'm not sure this is the recommended approach. As far as [I read](https://www.w3.org/TR/wai-aria-practices-1.2/#dialog_modal), we should focus the first focusable element:

image

In the example storybook, for instance, the input gets focused when you open the modal (which I think provides a great UX!)



> We add this listener

Not sure how this listener handles focus trapping? I mean, iterating (using TAB) through in-modal focusable elements. Would that listener make this test pass? https://github.com/holaluz/margarita/pull/303/files#diff-5007d823e5d98623f02c3fd45f67c9254171406ae3b2c52e9a2dae87add935eeR116-R163


> I mean complex in the sense that it introduces the concept of slot scopes

Hum, true that we're not using it yet, but it is an absolutely valid Vue core feature, and we might run into a circular logic trap here 😅  "we don't use slot scopes" -> "we don't introduce slot scopes because we're not using them" -> "we don't use slot scopes" 😂  

> In order to "close" the modal, we remove the listener and .focus() to the element that opened the modal (we can save its ref when opening).

Yep, this is what current implementation is doing, too 👍  
remove listener: https://github.com/holaluz/margarita/pull/303/files#diff-5007d823e5d98623f02c3fd45f67c9254171406ae3b2c52e9a2dae87add935eeR43-R55
focus back to trigger: https://github.com/holaluz/margarita/pull/303/files#diff-5007d823e5d98623f02c3fd45f67c9254171406ae3b2c52e9a2dae87add935eeR165-R173


> the DOM kinda gets a bit weird/hard to read at first glance when the button that opens the modal is inside the modal component itself.

(_I assume you meant the Vue tree, not the DOM, as the modal renders as a body direct child!_)

It is obviously a personal preference, but I'm not sure this is that _hard_ to read?

```vue
      <ma-modal title="title">
        <template #trigger="{openModal}">
          <ma-button @click="openModal">Open the modal</ma-button>
        </template>

        <template #content>
           <my-amazing-modal />
        </template>
    </ma-modal>
  ```


> Also what happens if we want a link in a text to open the modal. Do we have to set the whole text inside the #trigger? 

Hm, I think so 🤔 Or you could encapsulate the text-modal pack in a component, and make it render the clickable text


> What if we want to open it programmatically through js (IE to show an error)?

You can pass `openModal` as a prop to the child component, and there call it when needed 👍 


---

All that being said, I'm totally OK with exploring other alternatives such as having a global state to handle the modal if we see that it yields a better API and simplified implementation (which we should compare against modals with feature parity 😉).

"_We can (probably) instantiate it programmatically from inside the plugin and attach it to this.$root the first time we run $modal.open_" this sounds good – it is similar to what's being done here, with the difference that we're attaching a DOM node as a child of body.


Again, thanks for your thoughts! 🙌

**Comment by edufarre** [comment]
>> the DOM kinda gets a bit weird/hard to read at first glance when the button that opens the modal is inside the modal component itself.

> It is obviously a personal preference, but I'm not sure this is that hard to read?

In my opinion, it just seems a bit weird that the trigger of the modal is inside itself. HOWEVER, I do not believe it is hard to read nor understand, as the way it could be handled/implemented is quite flexible and straight forward.

>  mean complex in the sense that it introduces the concept of slot scopes, which we don't normally use (compared to plugins).

We do have scoped slots in `negroni's` project, and we also used them on the `nuxt-clean` repository as well. In this case, I believe @afontcu is right, we should not skip this approach because the concept is not widely used in our projects. This could be a great opportunity to start (or may I say continue) using them as they are a completely valid Vue feature.


> this.$once('hook:destroyed') and Abstract Components

Never seen this before in my life xD 🤷 

In a nutshell, I believe both approaches have their points. Nonetheless, I would probably try to use the `vue-simple-portal` lib, as having those components copy/pasted to our project might lead to a misunderstanding, that we should maintain them despite they do have a comment at the top, and having it wrapped into a `node_module` it might help (dunno, just my two cents here). @afontcu said that when trying to sue the lib, he bumped into some issues, dunno if the copy/paste approach would work for now, and we should iterate this in a future PR tho.

**Comment by afontcu** [comment]
I've pushed a commit that removes `this.$once('hook:destroyed')` to reduce the overall cognitive load  :)

**Comment by afontcu** [comment]
update: I was able to fix the duplicated Vue instances issue from `vue-simple-portal` 🎉  so a lot of magic is now gone

File: src/components/MaModal/MaModal.scss [hunk 1]
```diff
@@ -0,0 +1,82 @@
+.modal-overlay {
+  position: absolute;
```
Samitier, line 2: what happens when the user scrolls? Shouldn't the overlay be fixed too?

src/components/MaModal/MaModal.scss
```diff
@@ -0,0 +1,82 @@
...
+.modal-header {
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-start;
+  padding: get-space('xxsmall');
+}
+
+.modal-header--gradient {
+  background: linear-gradient(to right, get-color(orange), get-color(pink));
+  color: white;
```
Samitier, line 37:
```suggestion
  color: get-color(white);
```

afontcu: yeah, as mentioned in the PR post, I did not pay much attention into CSS details yet, and this is why the PR is marked as draft :)

src/components/MaModal/MaModal.scss
```diff
@@ -0,0 +1,82 @@
...
+.modal {
+  position: fixed;
+
+  // centering
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%, -50%);
+
+  // so that it doesn't reach end of viewport
+  width: 96vw;
```
Samitier, line 20: max-width maybe? What happens with responsive here?

src/components/MaModal/MaModal.scss
```diff
@@ -0,0 +1,82 @@
...
+
+.modal-content {
+  padding: 0 get-space('xxsmall');
+
+  // because fuck standards, that's why
+  // https://github.com/w3c/csswg-drafts/issues/129
+  margin-bottom: get-space('xxsmall');
+
+  // height management
+  max-height: 85vh;
```
Samitier, line 60: "hardcoding" the max height is a bit dangerous, no? For landscape devices it could add multiple vertical scrolls. Maybe we should make the content adapt the size of the modal and never make the modal bigger than the viewport (in kalimotxo it's done like that IIRC).

src/components/MaModal/MaModal.scss
afontcu: yep, nice catch. It should be `fixed`,  right? Fixing it up.

src/components/MaModal/MaModal.scss
afontcu: This is only meant to leave some gutter space at sides (so that the modal does not reach the beginning/end of the viewport). "Real" modal sizing is set here: https://github.com/holaluz/margarita/pull/303/files#diff-7e6f9b97a0c6c1a9b09d6cd5c7734c25f74239b4742386d2d361e50cf343068cR74-R82

I guess I could switch it up (set a max-width here, and then use the `width` prop to set its width. Does it make more sense?

src/components/MaModal/MaModal.scss
afontcu: > the content adapt the size of the modal

this is already happening

> and never make the modal bigger than the viewport

this is also happening, right? `85vh` means "85% of the height of the viewport".

Samitier:
Ai way better with the library. My brain says thanks ❤️ 

> re tabindex=-1 and focusing on close button, I'm not sure this is the recommended approach. As far as I read, we should focus the first focusable element:

From what I've read the tabindex <0  is to indicate we are setting the tab order in js. I think it applies here too.

> Not sure how this listener handles focus trapping?

Ok, my idea was to use the vanilla tab functionality and just restrict what elements can be focusable. So no listeners on keypress. Let me rephrase the snippet:

```js
// when we have opened the modal, we have focused in the close button 
// (or any focusable element inside the modal). 
// That means the focus is now inside a div called "modal"

// If we listen to this event, this will be called every time the user focuses in another element
// If they press tab or shift tab. Our objective is to never get the focus go away of the
// div modal
document.addEventListener("focus", event => {
  // if the modal is not opened or the focused element is inside the modal, we do nothing
 
 // but, if the modal is opened and the element we attempt to focus is outside the modal,
 // we "trap" the focus. That is, we don't focus on the "normal" element and we
 // focus on the first element inside the modal, to start the focus "loop" again.
  if (isModalOpen && !modalContainer.contains(event.target)) {
      event.stopPropagation()
      firstElementOfModal.focus()
  }
}, true)
```
Does it make sense? Maybe an executable example would be better 😅 

> Hum, true that we're not using it yet, but it is an absolutely valid Vue core feature.

Yes yes of course, we use them in dry too IIRC. But comparatively, we use plugins `$t`, `$layout`,  `$gtm` way way more.

Besides, aren't we pushing a bit far what slot scopes are originally for (slots in a v-for)? Why can't we just use refs for this? Why can't we just discard the trigger slot entirelly? I don't know, my sensation with this api is that we are adding way too much "noise" for something that should be trivial:

```vue
<a @click='modal.open'>Open modal</a>
<ma-modal #modal>
    modal content
</ma-modal>
```

> (I assume you meant the Vue tree, not the DOM, as the modal renders as a body direct child!)

I mean the trigger will be inside a `<div class="modal-wrapper"><div></div></div>`.

> You can pass openModal as a prop to the child component, and there call it when needed 👍

I mean what if I want to do it in JS? Do I really have to instantiate all the cases of a conditional modal in a template?

```js
sendCups() {
  // do stuff
  if (!isCupsValid) this.openModal(CupsErrorModal)
  else if (isGreatConsumer) this.openModal(GreatConsumerModal)
}
```

It's not a real use case but we could do that with a plugin 😬

src/components/MaModal/MaModal.scss
Samitier: I am improvising a bit, but why not:
```vue
// having this
<div class="modal-container">
  <div class="modal-overlay" />
  <div class="modal-content" />
</div>

// container and overlay occupy everything
.modal-container, .modal-overlay {
  position: fixed
  left: 0
  top: 0
  bottom: 0
  right: 0;
}

// container has a padding so that content always has 'spaces' in all sides
.modal-container {
   padding: 1rem;
 }

// content is simply a centered div
.modal-content {
   margin: auto;
   max-width: the size of the modal
}
```

src/components/MaModal/MaModal.scss
Samitier: yes but it's still "hardcoded" even if its a relative unit no? I mean it's 85% for any reason in particular? Shouldn't it be 100% (do not be bigger than your parent) and that's it?

Maybe I'm a bit lost so don't listen to me if it makes no sense 😂

afontcu:
> From what I've read the tabindex <0 is to indicate we are setting the tab order in js. I think it applies here too.

tabindex=-1 removes an element from the natural tab order, which is not want we want, here. Not sure if your approach will work, it looks good. What about shift+tab, though?

> Besides, aren't we pushing a bit far what slot scopes are originally for (slots in a v-for)? 

Hm not really, slot scopes are just a way to provide props to the slot content. There's not a forced relationship with iterations (even though they are used there)

> Why can't we just discard the trigger slot entirely? 

Because then we can't "extract" the `openModal` method. I wish we could, but that's [per design](https://vuejs.org/v2/guide/components-slots.html#Abbreviated-Syntax-for-Lone-Default-Slots). 

> I don't know, my sensation with this api is that we are adding way too much "noise" for something that should be trivial

Legit. I liked how everything was self-contained, but if people feel a plugin-based approach but yield a simpler, more comfortable API, I'm up for it 👍  What would it look like? Could you add a comment on the RFC?

> It's not a real use case

Troll alert 😂😂😂😂

src/components/MaModal/MaModal.scss
afontcu: It is totally hardcoded 😂 as hardcoded as the max widths, though

 it is not 100% because I wanted to let it breathe (instead of having the modal to fill the entire viewport height). This is a design decision after all, so happy to change it if design people think another value is more appropriate 👍

src/components/MaModal/MaModal.scss
afontcu: yeah, I think it looks good. I'll take a look 👍

Samitier:
> What about shift+tab, though?

It's the same, it will trigger the focus listener anyways. The only "weird" part is that it won't cycle backwards (starting from the bottom when we exit the focus from the top). When we arrive at the first focusable element of the modal it will stay there forever unless we add another if:

```js
if (isModalOpen && !modalContainer.contains(event.target)) {
      event.stopPropagation()
     if (is shift pressed) lastElementOfModal.focus()
     else firstElementOfModal.focus()
}
```

> Because then we can't "extract" the openModal method. 

I mean directly [accessing the component methods with a ref](https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements)  instead of a slot scope.

> Troll alert 😂😂😂😂

hahahah sorry

afontcu:
> unless we add another if

…and querying for focusable element… which will yield roughly the same focus management we're doing right now 😂 


> I mean directly [accessing the component methods with a ref](https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements) instead of a slot scope.

Oh, I see. Is that any better, though? If I'm not mistaken we'd be suggesting people to rely on implementation details.

Samitier:
> …and querying for focusable element… which will yield roughly the same focus management we're doing right now 😂

We could be clever and make the backdrop focusable (since it's clickable too). That way we have the close button as first focusable and the backdrop as last and we don't have to do a querySelectorAll (which is a bit fragile since it doesn't even consider disabled elements anyways). We would just need a ref and thats it. 

> Oh, I see. Is that any better, though? If I'm not mistaken we'd be suggesting people to rely on implementation details.

With TS this gets to another level of cool because you can expose methods as public and you have typing/autocompletion. So you can have components that expose APIs and still have private methods/props. But yeah, if the problem was that using scoped slots was uncommon in our code, accessing through $ref it's even weirder, I grant you that. So nevermind 😂

afontcu:
> We could be clever and make the backdrop focusable (since it's clickable too)

I think there's a fine line between cleverness and a11y issues 😂😂😂 

> (which is a bit fragile since it doesn't even consider disabled elements anyways)

yep this is true. We'd need to drop a `.filter(el => !el.hasAttribute('disabled'))` or something. I'll add this for the sake of completeness, even if we end up rewriting the focus logic to follow your approach 👍 



> But yeah, if the problem was that using scoped slots was uncommon in our code, accessing through $ref it's even weirder, I grant you that. So nevermind 😂

Yeah, let's drop this XD I'd say we either stick with the scoped slots approach or we explore the plugin way 😉. Let's see if anyone else wants to chime in? @holaluz/frontend

afontcu:
We could also simplify the focus management a bit by following this advice ([source](https://www.w3.org/TR/wai-aria-practices-1.2/#dialog_modal)):

>  If content is large enough that focusing the first interactive element could cause the beginning of content to scroll out of view, it is advisable to add tabindex="-1" to a static element at the top of the dialog, such as the dialog title or first paragraph, and initially focus that element. 

I'd drop a tabindex="-1" to the modal title, and always focus there on opening 🤷‍♀️

afontcu:
_re_ the debate between a slots-based API and a plugin-based one, we've come to the conclusion that slots are good enough even if the layout is a bit harder to grasp at the beginning because it allows to render "regular" components (with their props, events, and so on) as modal content, and because it does not prevent any common usage.

That being said, we need to figure out a way to toggle modal state programmatically (that is, from `<script>` section). We suggested some sort of prop (`v-model`, maybe?).

cc @Samitier @edufarre @alexcastelo95 (feel free to correct me if i'm wrong 😅)

alexcastelo95 (approved):

src/components/MaModal/MaModal.scss
```diff
@@ -0,0 +1,132 @@
...
+  background-color: var(--color-white-base);
+  word-break: break-word;
+}
+
+.modal-header {
+  display: flex;
+  justify-content: space-between;
+  align-items: flex-start;
+  padding: var(--spacing-medium);
+  gap: var(--spacing-medium);
```
Samitier, line 27: `gap` doesn't work on safari. Are we aware of that?

src/components/MaModal/MaModal.scss
```diff
@@ -0,0 +1,132 @@
+.modal-overlay {
```
Samitier, line 1: shouldn't we do it in css instead of scss?

afontcu: I was LITERALLY about to push a commit with this when I switched to github to read your comment 😂 ❤️

afontcu: https://github.com/holaluz/margarita/pull/303/commits/5a96327c59a6f9395fb9ee8bfe934d462045560f 😂

src/components/MaModal/MaModal.vue
```diff
@@ -0,0 +1,206 @@
...
+
+    flushQueue() {
+      return new Promise((resolve) => setTimeout(resolve, 0))
+    },
+
+    async setFocusableElements() {
+      await this.flushQueue()
+
+      // If we cannot find modal let's fail gracefully.
+      const modal = this.$refs['modal']
```
Samitier, line 172:
```suggestion
      const {modal} = this.$refs
```

src/scss/_transitions.scss
```diff
@@ -0,0 +1 @@
+@import 'transitions/slide-fade';
```
Samitier, line 1: I thought we said that transitions shouldn't be "global". Can't we scope this inside ma-modal.scss instead?

src/components/MaModal/MaModal.scss
Samitier: hahahah ❤️ ❤️ ❤️

src/components/MaModal/MaModal.vue
afontcu: done!

src/scss/_transitions.scss
afontcu: It was already exposed and I think it is being used in some places (will check). The only change here, which I'm not sure why happened, is that a `_transitions.scss` file was created instead of importing the transition from `_margarita.scss` 🤷‍♂️  

I'll revert this as it will yield some conflicts with #329 

We can then discuss if Margarita/DS should provide some transitions globally (at least they are only imported once, and not once per component, as it used to happen XD)

src/components/MaModal/MaModal.vue
```diff
@@ -0,0 +1,206 @@
...
+      if (!modal) return
+
+      this.focusableElements = modal.$el.querySelectorAll(FOCUSABLE_ELEMENTS)
+    },
+
+    handleTabKey(e) {
+      const firstElement = this.focusableElements[0]
+      const lastElement = this.focusableElements[
+        this.focusableElements.length - 1
+      ]
```
afontcu, line 184: Not gonna change it because it is hard to understand, but this could become a one-liner with a bit of gymnastics…

```js
const { length, [length - 1]: lastElement, 0: firstElement } = this.focusableElements
```

😬

src/scss/_transitions.scss
afontcu: done!

Samitier (approved):

src/components/MaModal/MaModal.css
```diff
@@ -0,0 +1,131 @@
...
+
+  background: linear-gradient(
+    to right,
+    var(--color-orange-base),
+    var(--color-pink-base)
+  );
+  color: var(--color-white-base);
+}
+
+.modal-header--gradient .modal-title {
```
edufarre, line 46: I know that [this PR](https://github.com/holaluz/margarita/pull/329) is still labelled as `not ready to merge`, but once it actually is, we could move this 'nesting' inside the definition of the actual `.modal-header--gradient` class, right?

Thus having something like:
```css
.modal-header--gradient{
  ...

  & .modal-title {
    text-align: center;
  }
}
```

Just saying this because it would be cool to bear it in mind to refactor this when applying the previously linked #329 PR

src/components/MaModal/MaModal.css
```diff
@@ -0,0 +1,131 @@
...
+.modal-transition-leave-active,
+.modal-transition-enter-active {
+  transition: opacity 0.3s ease;
+}
+
+.modal-transition-enter-active {
+  .modal {
+    animation: slideFromTop 0.4s ease-in-out;
+  }
+}
```
edufarre, line 117: I believe this is not ok 🤔 

```suggestion
.modal-transition-enter-active .modal {
    animation: slideFromTop 0.4s ease-in-out;
}
```

src/components/MaModal/MaModal.css
```diff
@@ -0,0 +1,131 @@
...
+.modal-transition-leave-active {
+  .modal {
+    animation: slideFromTop 0.4s ease-in-out reverse;
+  }
+}
```
edufarre, line 122: Same as previous comment

edufarre (approved):
Overall looks good to me. A quite easy component API, exhaustive tests thus being sure everything works as expected.

However, I kinda miss some comments in the component, this way we would be using our `vetur intellisense` feature, and have a way better DX when using it, as it's already in the project. I believe we should have to start using it 😄 

Great job guys! 🚀

src/components/MaModal/MaModal.scss
edufarre: Shall we try to tackle this before releasing a new version?🤔
I believe it is kinda blocker to use it in some of our front projects (as we do have traffic with `fasari` browser)

Samitier: I would use display: grid just in case for now

src/components/MaModal/MaModal.css
afontcu: hmmm sure, but why is it better?

afontcu:
> However, I kinda miss some comments in the component, this way we would be using our `vetur intellisense` feature, and have a way better DX when using it, as it's already in the project. I believe we should have to start using it 😄

Good point! Move widths array to an inline array so Vetur can pick it up. Not sure what additional comments would help, though – I believe all 3 props are self-explanatory?

src/components/MaModal/MaModal.css
afontcu: right! fixed 🚀

src/components/MaModal/MaModal.scss
afontcu: yep, agree. Didn't think about Safari (nobody does except for @Samitier 😂😂). I'll fix this. Thanks!!

afontcu: Handled it by setting a margin-left prop to the close icon instead of fully switch to grid :)

src/components/MaModal/MaModal.css
edufarre: It's just a matter of consistency when writing CSS across different components. No drama at all, I was thinking out loud

edufarre:
> > However, I kinda miss some comments in the component, this way we would be using our `vetur intellisense` feature, and have a way better DX when using it, as it's already in the project. I believe we should have to start using it 😄
> 
> Good point! Moved widths array to an inline array so Vetur can pick it up. Not sure what additional comments would help, though – I believe all 3 props are self-explanatory?

Yes, they are! I was just trying to make the point of using the `vetur intellisense` we already have integrated in the project, in order to improve our DX, sorry if I did not explain myself properly.

PR #307, opened 2020-12-19
fix(perf): Reduce build size by removing generated CSS from margarita-tokens

afontcu:
We had a CSS-generating file inside of margarita-tokens, so the output CSS was added **every time for every component**. Read that again: For. Every. Component!

I fixed it by porting back the animation to the only component using it. This resulted in some reduction in Margarita build (from 56.80 KiB to 48.60 KiB changing one line of code), but I'm also quite interested what's the outcome in bigger apps – again, we're currently adding this stupid CSS animation for every component. Every. Component.

From:

imatge

to:

imatge

Note: I also removed some other unused stuff such as a mixin, but it has no impact on build sizes.

_**edit**_: did a quick test. bonasera went from 127 KiB of CSS to 85 🚀


edufarre (approved):

Samitier:
yaayy! 😄 

Maybe it would be cool to convert all the global css to mixins if they are used in more than one place. That way they will only be added where we really use them.

Also, we could explore the extractCss option. Maybe if we don't have a global css at all and we keep it in the js bundle, tree shaking helps us get rid of css we don't use.

Samitier (approved):

afontcu:
> Maybe it would be cool to convert all the global css to mixins if they are used in more than one place. That way they will only be added where we really use them.

I checked, and there isn't much global CSS other than reset (from normalize) and other sensible defaults. We're good 👍 

> Also, we could explore the extractCss option. Maybe if we don't have a global css at all and we keep it in the js bundle, tree shaking helps us get rid of css we don't use.

Gave it a go, and I have mixed feelings:

* Parsing and executing a JS file is way more expensive than handling a CSS file, so reducing the amount of content in a JS file _makes sense_.
* OTOH, it would be great to tree-shake it, but I'm not sure this is even possible with UMD builds (remember that webpack is unable to compile stuff using ESM).

In the end, it all boil down to perf testing 😬 meausre baseline, then measure with extracting CSS, then compare.

PR #308, opened 2020-12-19
fix(perf): Remove support for dead browsers 

afontcu:
From this:

Captura de Pantalla 2020-12-19 a les 21 11 11

to this:

Captura de Pantalla 2020-12-19 a les 21 11 32

🎉 

10 KiB of CSS removed by excluding the following browers:

```
bb 10 (Blackberry Browser)
bb 7
ie 10 (Internet Explorer – obviously)
ie_mob 11 Internet Explorer Mobile (I didn't even know this was a thing)
ie_mob 10 (Opera Mobile)
op_mob 12.1
```

all browsers above are listed as dead ("dead" as in "_browsers without official support or updates for 24 months_"), so we should not be supporting them – even IE 10, which I believe we're already not supporting given the features we're currently using.

I suggest we created a shared browserslist config. Sadly enough Browserslist do not support `.js` config files so we can't simply add it on `npm-scripts`. Their docs suggest [creating a npm package](https://github.com/browserslist/browserslist#shareable-configs), but I'd like to stuff it into npm-scripts very much. Any ideas?

---

Notice that this PR along with #307 reduces our bundled CSS from ~60KiB to 38 by changing 2 lines of code. 💥

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,3 +1,3 @@
 > 1%
 last 2 versions
-not ie <= 8
+not dead


edufarre (approved):

Samitier:
> I suggest we created a shared browserslist config. Sadly enough Browserslist do not support .js config files so we can't simply add it on npm-scripts. Their docs suggest creating a npm package, but I'd like to stuff it into npm-scripts very much. Any ideas?

100% aggree because one of the goals of web warriors is to get real data about the browsers of the visitors we have and decide which browsers we support and which we don't. It could be perfect if we could use this effort to port it to the rest of our apps easily. Idea: maybe npm-scripts should be an npm monorrepo of plugins.

That is creating different npm packages from the same git repository, so we can do stuff like:

```
npm i @holaluz/npm-scripts/browserlist
npm i @holaluz/npm-scripts/eslint
etc
```

Here's some info on how we could implement it: https://blog.npmjs.org/post/186494959890/monorepos-and-npm

Samitier (approved):

afontcu:
> maybe npm-scripts should be an npm monorrepo of plugins

Yeah, that's a possible solution! Feels like an overkill for other stuff such as eslint or jest configs, and not sure what would mean if we want to move more responsibilities to npm-scripts, though.

Still, I say we merge this one, and then explore the better way to have a common browserslist config (which could be smarter than this, you know, with caniuse and analytics9

PR #309, opened 2020-12-21
fix: match media throws on Safari 

afontcu:
Even though docs mention that all browsers support `addEventListener` (source: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#Methods) turns out Safari does not. It only implements the deprecated `addListener` method, and throws when trying calling on the former.

imatge

this is currently happening on my Mac with Safari 13. 

Thus, we attempt to register the callback using `addEventListener`, but in case of failure we fall back to the deprecated `addListener`.

Related:
  https://github.com/mdn/sprints/issues/858
  https://stackoverflow.com/a/60000747/1042082


src/plugins/responsivePlugin.js
```diff
@@ -14,18 +14,15 @@ export default {
...
 
     updateCurrentBreakpoint()
 
     function updateCurrentBreakpoint() {
       const newBreakpoint = getCurrentBreakpoint(tabletQuery, desktopQuery)
 
-      if (newBreakpoint !== vue.prototype.$layout.currentBreakpoint) {
-        vue.prototype.$layout.currentBreakpoint = newBreakpoint
-      }
+      vue.prototype.$layout.currentBreakpoint = newBreakpoint
```
afontcu, line 25: Unrelated to the PR, but this check makes no sense

edufarre (approved):

PR #312, opened 2020-12-22
chore: Add tokens build when building storybook docs

edufarre:


package.json
```diff
@@ -27,7 +28,7 @@
     "test": "vue-cli-service test:unit",
     "test:watch": "vue-cli-service test:unit --watch",
     "tokens:build": "style-dictionary build --platform scss --config ./styleDictionary.config.js && prettier 'src/scss/variables/*.scss' --write",
-    "tokens:watch": "npm run tokens:build && chokidar  \"src/tokens/*.js\" -c \"npm run tokens:build\""
+    "tokens:watch": "chokidar  \"src/tokens/*.js\" -c \"npm run tokens:build\""
```
afontcu, line 31: As far as I can tell we're not supporting Windows here (notice that `format` uses single quotes) 🤷 , so we could clean this up a bit, too, and use single quotes

```suggestion
    "tokens:watch": "chokidar  'src/tokens/*.js' -c 'npm run tokens:build'"
```

package.json
```diff
@@ -16,6 +16,7 @@
   ],
   "scripts": {
     "prebuild": "npm run tokens:build",
+    "postinstall": "npm run tokens:build",
```
Samitier, line 19: is the prebuild necessary then?

afontcu: Hm, I think it is just in case someone changes a token without `tokens:watch` running?

edufarre: I believe Adri's right here! I would let it there, thus making sure the tokens we use are always up-to-date.

afontcu (approved):

PR #313, opened 2020-12-22
Add styles modal

alexcastelo95:
# Description
Implementation of styles to the 'MaModal' component of the PR https://github.com/holaluz/margarita/pull/303

**Task:** https://holaluz.kanbanize.com/ctrl_board/60/cards/12608/details/

# Application affected

- Margarita.

# Changes

- Apply modal DS styles.
- Apply Modal Maitai transitions.

# Discussion

I think we could treat as a new card or sub-task the implementation of the Modal without a header.

If you agree with me, the **Next steps** would be:

- [x] Implement a new story for the Multimedia Modal. (name totally invented).
- [ ] Implement tests.

Any comments are welcome. 

**pd:** thnks @carlotamolintr for your help.


src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,105 @@
...
+
+  &--pink {
+    background-image: url('../../assets/icons/icon-close-pink.svg');
+
+    &:hover,
+    &:active,
+    &:focus {
+      background-image: url('../../assets/icons/icon-close-darkpink.svg');
+    }
+  }
```
afontcu, line 80: since these are SVGs, could we use them as SVGs, and rely on `fill` or similar to change its color? We have the appropriate color variable, so loading two SVGs feels unnecessary here

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,105 @@
...
+  border-radius: 0.25rem;
+  box-shadow: 0 rem(2px) rem(8px) get-alpha-color(black, 0.67);
+  background-color: get-color(white);
   width: 96vw;
-
-  // temporary, stylistic stuff
-  background-color: white;
-  box-shadow: 0 25px 50px -12px rgba(0, 0, 0, 0.25);
-  border-radius: 0.5rem;
+  max-width: rem(500px);
```
afontcu, line 23: If nothing has changed, max-width should be set by the `width` prop

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,105 @@
...
 .modal-title {
-  font-weight: bold;
-  font-size: 1.3rem;
+  @include font-weight(bold);
+  display: flex;
+  flex: 1;
+  align-items: center;
+  margin: 0;
+  padding-right: 3rem;
+  min-height: 2.25rem;
```
afontcu, line 56: Do we really need a min-height, here?

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,105 @@
...
-  right: 0;
-  bottom: 0;
-
-  background-color: rgba(0, 0, 0, 0.25);
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  transition: opacity 0.4s ease;
+  z-index: 100;
```
afontcu, line 10: Do we need a z-index? We managed to handle the whole thing without indexes, I bet we could do the same after the changes, too

src/components/MaModal/MaModal.scss
alexcastelo95: You are right. I think there should be no problem. Now I'm researching how to do it. Thanks for the feedback.

src/components/MaModal/MaModal.scss
alexcastelo95: It is not necessary. You are right.

src/components/MaModal/MaModal.scss
alexcastelo95: No. With the padding we have enough.

src/components/MaModal/MaModal.scss
alexcastelo95: Well seen. I thought it was not positioned correctly, but without the z-index it maintains the behavior. Fixed!

src/components/MaModal/MaModal.scss
alexcastelo95: Solved!

src/scss/_margarita.scss
```diff
@@ -1,5 +1,5 @@
 @import 'margarita-tokens';
 @import 'normalize';
-@import 'transitions/slide-fade';
 @import 'utils/visually-hidden';
 @import 'ciutadella-loader';
+@import 'margarita-transitions';
```
afontcu, line 5: Could we keep the imports nominal, here? `margarita-transitions` sounds like something that we might export from the outside while it's not.


```
@import 'transitions/slide-fade';
@import 'transitions/fade';
```

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
...
-  border-radius: 0.5rem;
+  color: get-color(gray, dark);
 }
 
 .modal-header {
   display: flex;
   justify-content: space-between;
-  align-items: flex-start;
-  padding: get-space('xxsmall');
+  align-items: center;
```
afontcu, line 28: Close button should remaing on top even when title grows larger than one line:

```suggestion
  align-items: flex-start;
```

image

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
...
 
 .modal {
-  position: fixed;
-
-  // centering
-  left: 50%;
-  top: 50%;
-  transform: translate(-50%, -50%);
-
-  // so that it doesn't reach end of viewport
```
afontcu, line 19: Could we keep this comment? So people understands what's the point of this magic value 😇

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
   position: fixed;
   top: 0;
   left: 0;
-  right: 0;
-  bottom: 0;
-
-  background-color: rgba(0, 0, 0, 0.25);
+  display: flex;
+  flex-direction: column;
```
afontcu, line 6: I'd say this line is not needed

```suggestion
```

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
...
+  justify-content: center;
+  align-items: center;
+  transition: opacity 0.4s ease;
+  background-color: get-alpha-color(black, 0.5);
+  width: 100%;
+  height: 100%;
+  overflow-y: auto;
```
afontcu, line 13: Hum, what's the goal of this overflow? I'm not sure the overlay will ever overflow the viewport?

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
...
+  border-radius: 0.25rem;
+  box-shadow: 0 rem(2px) rem(8px) get-alpha-color(black, 0.67);
+  background-color: get-color(white);
   width: 96vw;
-
-  // temporary, stylistic stuff
-  background-color: white;
-  box-shadow: 0 25px 50px -12px rgba(0, 0, 0, 0.25);
-  border-radius: 0.5rem;
+  color: get-color(gray, dark);
```
afontcu, line 22: The Modal should not set the color of its inner elements. 

```suggestion
```

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
...
 }
 
 .modal-header--gradient .modal-title {
   flex: 1;
   text-align: center;
 }
 
+.modal-header--transparent {
+  display: none;
+}
```
afontcu, line 45: is this a leftover? 😄 

```suggestion
```

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
...
+
 .modal-title {
-  font-weight: bold;
-  font-size: 1.3rem;
+  @include font-weight(bold);
+  display: flex;
+  flex: 1;
+  align-items: center;
+  margin: 0;
+  padding-right: 3rem;
```
afontcu, line 53: Hm, this doesn't seem right, as modal title should be center aligned when header is Gradient. I might be missing something, but seems like removing all these lines makes it work?

```suggestion
```

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
...
+
+  @include mq($from: md) {
+    font-size: rem(24px);
+  }
```
afontcu, line 57: I would doublecheck this with design people, as I don't see it as part of the specification on Figma

src/components/MaModal/MaModal.scss
```diff
@@ -2,75 +2,94 @@
...
+  &--white {
+    color: get-color(white);
+
+    &:hover,
+    &:active,
+    &:focus {
+      opacity: 0.85;
+      color: get-color(gray, light);
+    }
+  }
```
afontcu, line 80: Could we doublecheck with design crew, too? Looks a bit off. (also, a bit of animation wouldn't hurt 😛)

src/components/MaModal/MaModal.scss
alexcastelo95: I think it is not necessary. Fixed ;)

src/components/MaModal/MaModal.scss
alexcastelo95: They were good intentions.

src/components/MaModal/MaModal.scss
alexcastelo95: Checked with design people!

src/components/MaModal/MaModal.scss
alexcastelo95: I'm not sure I understand what you're saying.
I think I deleted those styles and set the modal using flexbox. (do you want me to leave the comment there anyway? )

Thanks.

src/components/MaModal/MaModal.scss
alexcastelo95: Perfect. Sounds good to me.

src/components/MaModal/MaModal.scss
afontcu: Yeah, I mean the comment regarding the `width: 96vw` 😇

src/components/MaModal/MaModal.scss
```diff
@@ -2,41 +2,35 @@
...
+  border-radius: 0.25rem;
+  box-shadow: 0 rem(2px) rem(8px) get-alpha-color(black, 0.67);
+  background-color: get-color(white);
   width: 96vw;
-
-  // temporary, stylistic stuff
-  background-color: white;
-  box-shadow: 0 25px 50px -12px rgba(0, 0, 0, 0.25);
-  border-radius: 0.5rem;
+  word-break: break-word;
```
afontcu, line 20: This was a really nice catch 🙌

src/components/MaModal/MaModal.scss
```diff
@@ -45,32 +39,45 @@
 }
 
 .modal-title {
-  font-weight: bold;
-  font-size: 1.3rem;
+  @include font-weight(bold);
+
+  @include mq($from: md) {
+    font-size: 1.5rem;
+  }
```
afontcu, line 46: Just wondering if we verified this with the design crew? I'd say modals should always have the same font-size for titles, but they will know it better 😇

src/components/MaModal/MaModal.scss
afontcu: hahahahaha

src/components/MaModal/MaModal.scss
alexcastelo95: Oh sorry! Of course.

src/components/MaModal/MaModal.scss
alexcastelo95: Checked with DS Crew. The size for the `MaModal` title is:

- 24 pixels for all sizes.
- Bold size.

I have been told that when we have the `MaText` component it will correspond to the text size --> **Large**.

Fixed.

src/assets/icons/icon-close-pink.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="36" height="36"><defs><path d="M0 0h36v36H0V0zm1 1v34h34V1H1z" id="A"/><path d="M0 0h36v36H0V0zm1 1v34h34V1H1z" id="B"/></defs><g fill-rule="evenodd"><mask id="C" fill="#fff"><use xlink:href="#A"/></mask><g mask="url(#C)" fill="#e6007d" fill-rule="nonzero"><path d="M0 0h36v36H0z"/></g><mask id="D" fill="#fff"><use xlink:href="#B"/></mask><g fill="#e6007d" fill-rule="nonzero"><g mask="url(#D)" fill-opacity="0"><path d="M0 0h36v36H0z"/></g><path d="M10.864 9.205l15.9 15.9-1.818 1.818-15.9-15.9z"/><path d="M9.046 25.114l15.9-15.9 1.818 1.818-15.9 15.9z"/></g></g></svg>
```
afontcu, line 1: Can all these .svg files be removed, now?

src/components/MaModal/MaModal.scss
```diff
@@ -2,41 +2,37 @@
...
 .modal {
-  position: fixed;
-
-  // centering
-  left: 50%;
-  top: 50%;
-  transform: translate(-50%, -50%);
+  transition: transform 0.4s ease;
+  border-radius: 0.25rem;
+  box-shadow: 0 rem(2px) rem(8px) get-alpha-color(black, 0.67);
```
afontcu, line 17: Do we want borders that have a variable width depending on user settings of base font size? 😅 

```suggestion
  border-radius: 4px;
  box-shadow: 0 2px 8px get-alpha-color(black, 0.67);
```

(Until we have some border/shadow variables from the design system, I guess! But that's beyond the scope of this feature)

src/components/MaModal/MaModal.scss
```diff
@@ -2,41 +2,37 @@
...
+  word-break: break-word;
 }
 
 .modal-header {
   display: flex;
   justify-content: space-between;
   align-items: flex-start;
-  padding: get-space('xxsmall');
+  padding: get-space('xsmall');
+  border-radius: 0.25rem 0.25rem 0 0;
```
afontcu, line 30:
```suggestion
  border-radius: 4px 4px 0 0;
```

src/components/MaModal/MaModal.scss
```diff
@@ -2,41 +2,37 @@
...
 }
 
 .modal-header--gradient {
   background: linear-gradient(to right, get-color(orange), get-color(pink));
-  color: get-color(white);
-  border-top-left-radius: 0.5rem;
-  border-top-right-radius: 0.5rem;
+  color: white;
```
afontcu, line 35: Not that it matter much, but just for the sake of consistency:

```suggestion
  color: get-color(white);
```

src/components/MaModal/MaModal.vue
```diff
@@ -6,16 +6,15 @@
...
-        <div>
-          <div
-            class="modal-overlay"
-            data-testid="overlay"
-            @click="closeModal"
-          />
+        <div
+          class="modal-overlay"
+          data-testid="overlay"
+          @click.self="closeModal"
```
afontcu, line 16: Just double checking that the `.self` modifier is needed?

afontcu (approved):
Apart form the comments I left, it mostly looks good! 👍  I'd wait until someone else reviews it, too (you might need to nudge people 😉) and then merge it. After all this is not going straight to production, but rather merged into another PR.

good job!

src/components/MaModal/MaModal.vue
alexcastelo95: As the overlay is a container of the modal, if I remove the `.self` modifier, when I click on the modal, it closes.

I found this approach to solve this problem, although I am always open to suggestions. 😄

src/assets/icons/icon-close-pink.svg
alexcastelo95: Removed!

src/components/MaIcon/Icons/SquaredClose.vue
```diff
@@ -0,0 +1,24 @@
+<template>
```
Samitier, line 1: isn't the "squared close" just a close with a border? Do we need a separate svg for that?

src/components/MaModal/MaModal.vue
Samitier: shouldn't it be a separate div? I mean, not including the stack inside?

div modal
   div modal-overlay
   div modal content
/div

src/scss/transitions/fade.scss
```diff
@@ -0,0 +1,9 @@
+.fade-enter-active,
+.fade-leave-active {
+  transition: opacity 0.3s cubic-bezier(0.4, 0, 1, 1);
+}
+
+.fade-enter,
+.fade-leave-to {
+  opacity: 0;
+}
```
Samitier, line 9: I thought this global animation css died. Can we scope the animation where it's used?

src/components/MaIcon/Icons/SquaredClose.vue
alexcastelo95: Yes, it is. I understand you're saying that because we could manage it with just one icon, right?

Here I let myself advise by your experience. We have 2 options:

- Have only the icon "close" and work in each case with the border of the <div> where it is used. 
- Have the component of the "square close" as a reusable icon.

I find it difficult to see the pros and cons of one and the other. Maybe it's a question of size (svg is always heavier than applying a class to a div)?

Thanks for the feedback.

src/scss/transitions/fade.scss
alexcastelo95: I think we used it when I started the task of the **modal styles**, but in the end we have applied the transition scoped in the component.

Deleted.

src/components/MaModal/MaModal.vue
alexcastelo95: I don't know. I use the overlay to position the Modal.

Must be a separate div? I'm open to suggestions.

afontcu: Yeah, I would have the overlay as a sibling node of the modal content (I think (think!) it is the initial approach I followed)

src/components/MaIcon/Icons/SquaredClose.vue
afontcu: If possible, I'd also keep it simple using a single SVG and use borders.

alexcastelo95: Done!

src/components/MaModal/MaModal.vue
alexcastelo95: If I do it this way I can't get the `ma-stack` right (it doesn't appear).

I think I'm not understanding well how the `openModal` works...

Blai, if you have some time, we could do it in pairing? 👀 

Thanks in advance.

alexcastelo95: I have found a problem with that approach, a transition cannot encapsulate 2 elements. Therefore I have created a wrapper for the `modal-content` and the `modal-overlay`.

It is not the cleanest solution, and I am using the z-index.

Next week I'll look at it calmly.

afontcu: > I have found a problem with that approach, a transition cannot encapsulate 2 elements

A `<transition-group>` can, though: https://vuejs.org/v2/api/#transition-group

edufarre (approved):

PR #315, opened 2020-12-24
feat: Remove MaGrid

### LINKED ISSUE #314: Deprecate and remove current MaGrid
Current MaGrid(/Container/Column/Row) implementation is flawed and offers a poor DX. Thus, I suggest we remove it entirely from Margarita even before we start adding a new grid/columns system.

The suggested removal path is to installit from [margarita-deprecated](https://github.com/holaluz/margarita-deprecated), rename them to something local (MaGridContainer -> BonaseraGridContainer), and then erase them from Margarita (by releasing a major version).

- [ ] Remove them from bonasera (https://github.com/holaluz/bonasera/pull/773).
- [ ] Remove them from maitai (https://github.com/holaluz/maitai/pull/363).
- [ ] Remove them from dry-martini (https://github.com/holaluz/dry-martini/pull/389).
- [ ] Remove them from one-click-sales (_add here a link to the PR when opened_).
- [ ] Remove them from margarita (https://github.com/holaluz/margarita/pull/315).

afontcu:
Closes #314

I just opened the PR, it is not ready to merge until we agree on a way to share the deprecated MaGrid components 😉.

---

Suggested steps:

- [x] Move MaGrid components from margarita to https://github.com/holaluz/margarita-deprecated/
- [x] Replace import from margarita to margarita-deprecated in projects where MaGrid components are used (bonasera, dry, maitai, and oneclicksales). I'd use something similar to https://github.com/holaluz/margarita/issues/314#issuecomment-750841127
- [ ] Merge this PR (will release a new major version)


Samitier:
Can we just grab the transpiled grid.scss (and the css of the deleted components, which is already global) and just leave it as a legacy/optional css in margarita?

Something like:

```js
  css: [
    '@holaluz/margarita/dist/legacy-grid.css',  // here
    '@holaluz/margarita/dist/margarita.css',
    '@holaluz/kalimotxo/dist/kalimotxo.css',
  ],
```

Which will add the classes `.ma-grid-container` `.ma-grid-row`, columns, etc. The components are simply css classes and a div, so porting that to pure css seems pretty easy.

sergigraciaj:
> Can we just grab the transpiled grid.scss (and the css of the deleted components, which is already global) and just leave it as a legacy/optional css in margarita?
> 
> Something like:
> 
> ```js
>   css: [
>     '@holaluz/margarita/dist/legacy-grid.css',  // here
>     '@holaluz/margarita/dist/margarita.css',
>     '@holaluz/kalimotxo/dist/kalimotxo.css',
>   ],
> ```
> 
> Which will add the classes `.ma-grid-container` `.ma-grid-row`, columns, etc. The components are simply css classes and a div, so porting that to pure css seems pretty easy.

I think having a repo for the deprecated components is a better approach.
As these components have props, I guess having just classes will make the breaking change a little more complex than just a search&replace as we have to do having a new repo with the deprecated classes.

afontcu:
> Can we just grab the transpiled grid.scss (and the css of the deleted components, which is already global) and just leave it as a legacy/optional css in margarita?

yeah, thought about that. We could even add a `legacy: true` config option in Margarita to opt-in to the deprecated styles.

That being said, isolating all the cruft from these components to its own repo and leave Margarita unaware of them felt "cleaner" 🤔

Samitier:
After "playing" for a bit with the idea of not having all this components in this PR -> https://github.com/holaluz/maitai/pull/364

I still think we need a `ma-container`, which it's implementation could be something way more trivial than the abomination it used to be. Something like:

```vue
<template>
    <div class="ma-container">
       <slot/>
    </div>
</template>

<style scoped>
.ma-container {
  position: relative;
  margin: auto;
  max-width: 1180px;
  padding: 0 1rem;
}
</style>
```

I don't know if this is the PR to do this but I kinda not like the idea of removing features of a repo without leaving some kind of non deprecated alternative. Also, IMHO a container has nothing to do with grids. I think the concept was mixed when ma-grid was implemented and staid that way and we are not even discussing it in the issues or the DS.

But regardless of that, LGTM!

Samitier (approved):

afontcu:
> Also, IMHO a container has nothing to do with grids

Totally agree. This is something we def need to tackle.

> I kinda not like the idea of removing features of a repo without leaving some kind of non deprecated alternative

The thing here is that MaGridContainer is deprecated, but not removed, as it is still available through margarita-deprecated, so an immediate alternative is not needed (it is desired, though!).

What if we left things as they are in the PR right now, and we discuss a Container/Wrapper implementation soon enough? The implementation you mention is simple enough that we could easily replace local implementations with the one from Margarita.

Does it sound good?

PR #316, opened 2020-12-28
feat: Change layout spacing names and values

alexcastelo95:
I have implemented the new spacing in **Margarita**. I have reviewed the implementation and it seems to be working as agreed.

The only thing that screams to me is that the function of **spacing.js**, the one that generates, the spacing variables in **_spacing.scss,** adds me by default a dash between those variables that begin with a number (2xlarge, 3xlarge, ....).

Any suggestion is welcome. 👍🏼


afontcu:
> The only thing that screams to me is that the function of **spacing.js**, the one that generates, the spacing variables in **_spacing.scss,** adds me by default a dash between those variables that begin with a number (2xlarge, 3xlarge, ....).

Not sure I follow: you mean you cannot set `'2xlarge': '4rem',` in `spacing.js`?

---

oh, btw, this is a breaking change. A small one, because Stack is only used in a handful of places, but it might require some tweaks.

alexcastelo95:
Sorry, I explained it wrong. The function spacing.js generates the tokens of `_spacing.scss`. In cases where the token starts with a number, the following happens.

(spacing.js) **2xlarge** -------> **2-xlarge** (_spacing.scss)

In principle, I have checked where `MaStack` is used in **Margarita**. If `MaStack` is used in other projects it will surely be affected, since we have changed the nomenclature of sizes.

I can leave it for us to check it with you when you come back.

alexcastelo95:
I've checked a little. For now, there are certain things that make me doubt:

- id's in CSS I think you can't start with numbers, with classes you can. (but I think that's not the problem).

- From what we saw with Edu, it seems to be a Style Dictionary problem, which by default adds hyphens between any non-text character.

My proposal is, let's look until Wednesday if there is any other way to manage these variables.

If not, let's take the following nomenclature for the major spacing tokens as a guide:

- 2-xl
- 3-xl
- 4-xl
- 5-xl
- 6-xl

What do you think about it? 🤔

afontcu:
>     * From what we saw with Edu, it seems to be a Style Dictionary problem, which by default adds hyphens between any non-text character.

It is due to lodash' `kebabCase`. `_.kebabCase('3abc') === '3-abc'`  ([source](https://github.com/amzn/style-dictionary/blob/master/lib/common/transforms.js#L201))

We could either:
1) Find a different naming
2) Create a custom transformer to override style-dictionary's [name/cti/kebab](https://amzn.github.io/style-dictionary/#/transforms?id=namectikebab) one

icurdu:
Changing to kebab-case seems to be the best approach in my opinion.

PR #317, opened 2021-01-11
feat: Autogenerate component & prop autocompletion for Margarita components

afontcu:
This PR adds a first iteration of the component autocompletion & intellisense that Vetur can offer.

## What does the PR do

It yields the following result:

Captura de Pantalla 2021-01-11 a les 11 42 16

imatge

## What's missing

It doesn't add "values" or description for each prop. This needs to be done manually by adding JSDoc comments to each prop. [Here](https://github.com/CyCraft/vue-intellisense/blob/production/packages/scripts/test/helpers/BlitzForm.vue#L9) you have an example of available fields.

btw, these JSDoc comments will also be used by Storybook or any other platform to document components, so work done there will be used everywhere and will become the source of truth for props documentation. Quite neat.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,6 @@
   ],
   "scripts": {
     "prepare": "npm run tokens:build && npm run vetur:build",
-    "prebuild": "npm run tokens:build",
     "build": "vue-cli-service build --target lib --name margarita src/index.js",
     "format": "prettier 'src/**/*.{js,vue}' --write && stylelint 'src/**/*.scss' --fix",
     "lint": "npm run lint:js && npm run lint:css",
--- a/package.json
+++ b/package.json
@@ -12,23 +12,25 @@
   "files": [
     "dist",
     "src",
-    "types/*.d.ts"
+    "types/*.d.ts",
+    "vetur"
   ],
   "scripts": {
+    "prepare": "npm run tokens:build && npm run vetur:build",
     "prebuild": "npm run tokens:build",
     "build": "vue-cli-service build --target lib --name margarita src/index.js",
     "format": "prettier 'src/**/*.{js,vue}' --write && stylelint 'src/**/*.scss' --fix",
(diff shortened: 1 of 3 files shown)


afontcu:
By the way, this is what files in `/vetur` look like:

Captura de Pantalla 2021-01-11 a les 12 09 15

Captura de Pantalla 2021-01-11 a les 12 09 27

package.json
```diff
@@ -12,11 +12,12 @@
...
     "dist",
     "src",
-    "types/*.d.ts"
+    "types/*.d.ts",
+    "vetur"
   ],
   "scripts": {
-    "postinstall": "npm run tokens:build",
-    "prebuild": "npm run tokens:build",
+    "postinstall": "npm run tokens:build && npm run vetur:build",
```
afontcu, line 19: So that we have a better DX when developing Margarita locally, too.

Samitier:
Amazing 🤩 

One thing not entirely related to this: do you think we should make our projects open a prompt in vscode with vetur as a "workspace recommendation"? Because it's kinda already a prerequisite to be able to code correctly (also vetur, eslint, prettier, etc).

https://tattoocoder.com/recommending-vscode-extensions-within-your-open-source-projects/

Not sure where to put it to share it between projects, though.

Samitier (approved):

afontcu:
> One thing not entirely related to this: do you think we should make our projects open a prompt in vscode with vetur as a "workspace recommendation"? 

Sure! We actually have it in [bonasera](https://github.com/holaluz/bonasera/tree/master/.vscode). Not sure it is updated.

> Not sure where to put it to share it between projects, though.

Hm, it would be quite cool if we could extend recommendations, so a base file could live in npm-scripts. Let me check it out.

edufarre:
Looks promising! Can't wait for it!

edufarre (approved):

Issue #305, opened 2020-12-18
How should we treat and handle shadows?


edufarre:
## Current approach

We have them implemented as scss variables:
```scss
$shadow-light: rgba(0, 0, 0, 0.027451);
$shadow-medium: rgba(0, 0, 0, 0.0980392);
$shadow-dark: rgba(0, 0, 0, 0.498);
```

## Suggested approach (by @Samitier ) commented on [this PR](https://github.com/holaluz/margarita/pull/302#discussion_r545636122)

They are not actually shadows per se, just blacks with opacity. Maybe we should just unify the entire box-shadow in a variable or treat the shadow color as just another color.

I mean this as box shadow:
```scss
$shadow-light: npx npx npx rgba(0, 0, 0, 0.027451);
```
Or this as color:
```scss
shadow-base: rgba(0, 0, 0, 0.027451);
shadow-light: ...
shadow-dark: ...
```

icurdu:
Most of our shadows are black with opacity, but there's one type of shadow that we use for images on the homepage that has another color, and we want to use it more often. 

After some discussion in design, we have an [initial approach to shadows](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/Design-System-Beta?node-id=1739%3A0), based on where are these used. We wanted to have opacity and color as variables, but maybe it's better to have strict values. 

As I'm writing this, one question came out: Should we name the shadows following the clothing sizing convention?

afontcu:
> Should we name the shadows following the clothing sizing convention

As far as I know (correct me if I'm wrong) shadows are used as "layers" of depth. With that in mind, I'm not entirely sure that naming shadows as clothing sizes would convey that meaning, but I'm also unable to provide an alternative 😂

edufarre:
> As I'm writing this, one question came out: Should we name the shadows following the clothing sizing convention?

I would not use 'clothe sizing' nomenclature in this case, that would be weird when talking about shadows, isn't it?

I believe we should use the same approach as our DS colors:
```ts
shadowTypes: 'darker' | 'dark' | 'base' | 'light' | 'lighter' 
```

Samitier:
Maybe a quick search through our repositories could tell if our shadows are consistent and need to be a variable. For example:

In dry we have more or less 9 different shadows (but I think that a lot of them are the same shadow written differently). The only shadows that are truly unique are white shadows overlying images and the brown shadow that @icurdu says.

[...]

PR #318, opened 2021-01-11
chore: Port missing tokens to style-dictionary

afontcu:
This PR doesn't add any feature, it simply ports missing tokens (shadows, breakpoints…) to style-dictionary and cleans up its folder structure a bit.

We also changed the output SCSS to make it easier to eventually add CSS custom properties (incoming PR).

Notice I also moved style-dictionary stuff into `.style-dictionary/` to mimic storybook and other "helper" utilities that are not "source files" of the UI library.

Also notice the PR is not a breaking change because get-color and get-space did not change their behavior, and that we're still exposing shadow vars:

```scss
$shadow-light: rgba(0, 0, 0, 0.027451) !default;
$shadow-medium: rgba(0, 0, 0, 0.0980392) !default;
$shadow-dark: rgba(0, 0, 0, 0.498) !default;
```


.style-dictionary/transformToStyleDictionary.spec.js
```diff
@@ -0,0 +1,29 @@
+import transformToStyleDictionary from './transformToStyleDictionary'
```
Samitier, line 1: if `.style-dictionary` dir is autogenerated, shouldn't we have this test somewhere else?

Samitier: nevermind, I didn't understand the PR 😂

Samitier:
maybe it's a good moment to re think if this shadow variables are really useful or not. Why don't they have the rest of the box-shadow props (not only the color)?

Samitier (approved):

alexcastelo95 (approved):

afontcu:
> maybe it's a good moment to re think if this shadow variables are really useful or not

I wanted to keep this PR non-breaking so we could merge it without any consequences. I'm not 100% sure if shadow variables are being used and where (I assume they are), so we might need to investigate a bit before removing them 👍 

That being said, I don't care that much about current shadows 😂 they are not being used consistently and do not convey the expected meaning ("layers" of depth), so… we really need to talk with design folks about getting them right (#305)

PR #322, opened 2021-01-12
docs: Replace deprecated space naming by the new valid one

edufarre:


afontcu (approved):
heads up! It is a docs change, so no need to issue a release.

`docs: Replace deprecated…`  will do ;)

alexcastelo95 (approved):

PR #323, opened 2021-01-13
fix: Remove rem function usage

afontcu:
On the quest to progressively reduce the usage of SCSS, this PR removes the dreaded `rem()` function, which was not that helpful.

Notice that the function is still exposed for consumers to use, so this is not a breaking change.

Also, I tweaked some values to make them closer to what's in the Figma file of the Design System, so I think we could issue a release.

### DIFF
--- a/src/components/MaAlert/MaAlert.scss
+++ b/src/components/MaAlert/MaAlert.scss
@@ -3,30 +3,30 @@
   color: get-color(gray, darker);
 
   &--small {
-    padding: rem(4px);
+    padding: 0.25rem;
 
     @include mq($from: sm) {
-      padding: rem(8px);
+      padding: 0.5rem;
     }
 
     .alert-banner__icon {
-      margin-right: rem(8px);
+      margin-right: 0.5rem;
     }
   }
 
   &--medium {
-    padding: rem(8px);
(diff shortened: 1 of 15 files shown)


sergigraciaj (approved):

sergigraciaj:
Why don't we remove the rem function and make it a breaking change? The investment isn't worth it?

Samitier:
> Why don't we remove the rem function and make it a breaking change? The investment isn't worth it?

I agree with @sergigraciaj, let's burn it. Worth the investment for a progressive move to postcss and we will make sure all of our pages are made with rems in mind, not px.

afontcu:
Agree on burning it! However, I thought about grouping several breaking changes regarding SCSS and releasing them together (`9.x` probably). For instance, `get-color`, `font-weight`, and even `gradient-image` could go away, and potentially all SCSS variables (replaced by CSS custom props). We'd need to talk about `sass-mq`, too.

Since every breaking change requires some work on our side, I thought doing it once would be sensible.

Do you think it makes sense to issue a single release, or you'd rather have several ones, to make it more clear? // @sergigraciaj @Samitier (btw, I love your willingness to burnt anything SCSS-related XD)

PR #324, opened 2021-01-15
Review modal styles

alexcastelo95:
I'm not getting the `modal` and `overlay` transitions to work at different times.

We have looked at [Vue's ](https://vuejs.org/v2/guide/transitions.html)docu but it is not working. We also tried to make a `transition-group`, but it didn't work either).

If anyone wants to try it, I'll upload the changes. I will keep trying today.

### DIFF
--- a/src/components/MaModal/MaModal.scss
+++ b/src/components/MaModal/MaModal.scss
@@ -8,17 +8,15 @@
 }
 
 .modal {
+  position: fixed;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 96vw; // so that it doesn't reach end of viewport
+
   border-radius: 4px;
   box-shadow: 0 2px 8px var(--shadow-dark);
   background-color: var(--color-white-base);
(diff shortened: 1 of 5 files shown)


src/components/MaModal/MaModal.scss
```diff
@@ -1,29 +1,22 @@
...
-
 .modal-overlay {
   position: fixed;
-  transition: opacity 0.2s ease;
   background-color: get-alpha-color(black, 0.5);
   width: 100%;
   height: 100%;
+  top: 0;
+  left: 0;
+  z-index: 1000;
```
afontcu, line 8: Do we actually need z-indexes, here and in `.modal`?

alexcastelo95: Deleted. It is working 🚀

afontcu:
This looks quite solid, I'll go ahead and merge this one onto the main Modal PR so we can review the whole thing!

PR #325, opened 2021-01-15
chore: Remove font weight usage

alexcastelo95:


src/scss/_margarita-tokens.scss
```diff
@@ -7,7 +7,6 @@ $rem-baseline: 16px;
 @import 'functions/spacing';
 
 @import 'mixins/font-mixin';
-@import 'mixins/font-weight';
```
afontcu, line 10: heads up!! we shouldn't remove `font-weight` just yet… otherwise we'd be introducing a breaking change. As discussed in last FE crew meeting and [here](https://github.com/holaluz/margarita/pull/323#issuecomment-759438993), let's first remove all margarita usages and finally issue a single release with all breaking changes related to SCSS utilities.

afontcu:
The PR (and thus, the merge commit) should not be a `fix:` but rather `chore:` or something similar that doesn't trigger a release

src/scss/_margarita-tokens.scss
alexcastelo95: Sorry. Then I didn't quite understand the procedure of how to delete it.

I will leave the functions of the scss and the mixin as they were. I understand that I only have to remove from Margarita those places where this one is used.

afontcu: > I understand that I only have to remove from Margarita those places where this one is used.

Yeah! Exactly. So PRs do not become a breaking change until we set up a new PR removing all of them, and then releasing a new version :)

alexcastelo95: Okay, now I get it. Thank you for your feedback and for your patience. 😃

Samitier (approved):

src/scss/_margarita-tokens.scss
afontcu: No problem :) Make sure you add it back! 😉

alexcastelo95:
I have left the mixin included, I think everything is in order now.

afontcu (approved):

edufarre (approved):

PR #326, opened 2021-01-15
chore: Remove SCSS get functions

afontcu:
- Remove `get-space`
- Remove  `get-color`
- Clean up storybook colors page a bit

btw, following is the regex I used in VSC to replace all colors, which will come in handy when removing them from projects:

```
search: get-color\((.*\w)\)
replace: var(--color-$1-base)

search: get-color\((.*\w), (.*\w)\)
replace: var(--color-$1-$2)
```


Samitier (approved):
🤩

src/components/MaStack/MaStack.vue
```diff
@@ -37,12 +37,35 @@ export default {
...
     classes() {
       return {
-        [`stack--space-${this.responsiveSpace}`]: true,
         [`stack--align-${this.responsiveAlign}`]: this.align,
       }
     },
+
+    styles() {
+      return {
+        '--stack-gap': spacing[this.responsiveSpace],
```
afontcu, line 46: This looks cool, but I'm a bit worried it might pollute the generated HTML **a lot**. Maybe we should simply create a class for each spacing?

edufarre (approved):

src/components/MaStack/MaStack.vue
Samitier: please dont kill me but why not simply

```vue
<div :style="{gap}">...

computed:{
 gap() { return spacing[this.responsiveSpace] }
}
```
 We will never have the need of overwriting this from outside this component (and if we do, we have the prop) so honestly what is the downside of an inline style besides being a bit ugly? 😂 
 
 I can already feel the guns pointing at me hahahah

afontcu: Actually… I don't see any benefit about using a local custom property over the inline styling 😂 maybe the specificity of the resulting selector, but we could see that as a feature, not as a bug 😂

afontcu: done!

PR #327, opened 2021-01-18
feat: Add MaText component

### LINKED ISSUE #296: [RFC] MaText component
- Start Date: 09/12/2020
- Target Minor Version: 6.2.0
- Related to: #255
- Implementation PR: #295

# Summary

Add the typography rules from the Design System to a `MaText` Margarita component.

# Basic example

Using component slot:
```vue
<ma-text size="xsmall" tag="p">
  Hello everybody
</ma-text>
```
Using `v-text` directive:
```vue
<ma-text size="xsmall" tag="p" v-text="'Hello everybody'"/>
```

Using our `v-markdown` directive:
```vue
<ma-text size="xsmall" tag="p" v-markdown="'Hello **everybody**'">
```

# Motivation

This issue aims to resolve the inconsistency between font styles across the different device's screen sizes .

# Component API

The component could have to props; `tag` and `size`. Where `tag` could be one of the text tags that we use the most:
```js
tag: 'p', 'span', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' // span as default?
```
and `size` one of the agreed size names on the DS:
```js
size: 'xsmall', 'small', 'base', 'medium', 'large', 'xlarge', 'xxlarge',  'xxxlarge' // base as default?
```

# Alternatives

We could use utility classes instead of a component as already commented on the related issue on the first section.

```vue
<span class="xsmall">
  Hello everybody
</span>
```

# Adoption strategy

This is not a breaking change, thus allowing us start using this component at the same time that replacing the different text tags by this new implemented component.

# Unresolved questions

### How should be approach the line height's negative margin?
* https://medium.com/eightshapes-llc/cropping-away-negative-impacts-of-line-height-84d744e016ce
* https://medium.com/eightshapes-llc/space-in-design-systems-188bcbae0d62#9406

A good option here would be using the suggested tool in order to compute the negative margins that a speceific `line-height` should have. [Link to text crop tool](http://text-crop.eightshapes.com/). Once knowing them we could add these to our project's tokens, thus tackling the targeted issue.

### What should we do with the different font families?
* Shall we add `Canela` font family as well to the project?
If so, we should bear in mind that that would require a new component's `prop`, which allowed values would be:
```js
font: 'Ciutadella', 'Canela', 'Arial' // Ciutadella as default?
```
finally being something like:
```vue
<ma-text size="xsmall" tag="p" font="Canela">
  Hello everybody
</ma-text>
```

edufarre:
## Still pending to decide if the `tone` tokens naming is appropriate:
```js
  white: color.white.base,
  red: color.red.base,
  pink: color.pink.base,
  blue: color.blue.base,
  green: color.green.base,
  yellow: color.yellow.base,
  'gray-darker': color.gray.darker,
  'gray-dark': color.gray.dark,
  gray: color.gray.base,
```
and then set `gray-dark` tone as `default` 

## I also added some comments in order to use the vetur intellisense:

Component Props:
Screen Shot 2021-01-26 at 09 54 19

Prop values:
Screen Shot 2021-01-26 at 09 54 29

Closes #296


edufarre:
Pending to discuss:
https://github.com/holaluz/margarita/issues/296#issuecomment-762323799

src/components/MaText/MaText.spec.js
```diff
@@ -0,0 +1,90 @@
...
+
+        expect(getByText(slotText)).toHaveStyle(
+          `color: var(--color-${desiredColor})`
+        )
+      }
+    )
+  })
+})
+
+const renderComponent = (defaultText = 'Hello everybody', props = {}) => {
```
afontcu, line 78: The function sets a `defaultText` but then it always provide a `slotText` variable xD I think we could ignore this param and make the signature simpler:

`function renderComponent(props) { ... }`

src/components/MaText/MaText.stories.js
```diff
@@ -0,0 +1,51 @@
...
+import { colorNamesList } from '@margarita/utils/colorNamesList'
+
+const { textSize, textTags } = TEXT_TOKENS
+const colorValues = colorNamesList
+
+export default {
+  title: 'Components/Text',
+}
+
+export const Text = () => {
```
afontcu, line 13: Maybe we could add stories that showed all the variants, instead of making them dynamic, as we did in the [Alert](https://holaluz.github.io/margarita/?path=/story/components-alert--alert) or the [Pill](https://holaluz.github.io/margarita/?path=/story/components-pill--pill). This would allow people to compare sizes visually, for example.

src/components/MaText/MaText.vue
```diff
@@ -0,0 +1,70 @@
...
+
+    color: {
+      type: String,
+      default: 'gray-dark',
+      validator: (val) => {
+        const composedColor = val.split('-')
+        const targetColor = composedColor[0]
+        const targetTone = composedColor[1] ? composedColor[1] : 'base'
+
+        return colorNamesList.includes(`${targetColor}-${targetTone}`)
```
afontcu, line 39: the following would work, too, but it's just a stylistic suggestion, albeit shorter and terser. feel free to ignore it xD


```js
const [color, tone] = val.split('-')
return colorNamesList.includes(`${color}-${tone || 'base'}`)
```

afontcu (commented):
Hey! The general direction looks great 💪💪  I left some comments here and there!

src/tokens/text.js
```diff
@@ -0,0 +1,104 @@
...
+  '3-xlarge': {
+    'font-size': '2.5rem',
+    'line-height': 1.1,
+    'font-weight': 'bold',
+  },
+  '4-xlarge': {
+    'font-size': '2.5rem',
+    'line-height': 1.1,
+    'font-weight': 'bold',
+  },
```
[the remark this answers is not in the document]
edufarre: They are the same on `mobile` and `tablet` breakpoints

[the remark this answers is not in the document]
afontcu: cool! resolving comment

src/utils/colorNamesList.spec.js
```diff
@@ -0,0 +1,12 @@
+import { colorNamesList } from './colorNamesList'
+import { color } from '@margarita/tokens'
+
+describe('colorNamesList', () => {
+  test('generates a list with all the tokenized colors', () => {
+    Object.entries(color).forEach(([colorName, tones]) => {
+      Object.keys(tones).forEach((toneName) => {
+        expect(colorNamesList.includes(`${colorName}-${toneName}`)).toBe(true)
```
afontcu, line 8: I feel it's kinda weird to have a test for a variable 🤔 what about turning it into a function? Then the test could use a simple, demo object, that would help understand the function.

src/components/MaText/MaText.spec.js
```diff
@@ -0,0 +1,90 @@
...
+
+describe('MaText', () => {
+  const slotText = 'Test text'
+  test('renders content from slot', () => {
+    const { getByText } = renderComponent(slotText)
+
+    expect(getByText(slotText)).toBeInTheDocument()
+  })
+
+  describe('Property validators tests', () => {
```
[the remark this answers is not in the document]
edufarre: I believe it might be better to create a `colorPropValidator` and thus being able to use it across different components. However, I do think that it is nice to have these functions tests, in the end, our property validation functions give some behaviour logic to our component, and having the test that asserts that the validation is working as expected, it is kinda cool, although it might be a bit overkill?

src/components/MaText/MaText.spec.js
```diff
@@ -0,0 +1,64 @@
...
+const slotText = 'Hello everybody'
+
+describe('MaText', () => {
+  test('renders content from slot', () => {
+    const { getByText } = renderComponent()
+
+    expect(getByText(slotText)).toBeInTheDocument()
+  })
+
+  describe(`'Tag' property test`, () => {
```
afontcu, line 17: I think these `describe` block are a bit redundant because there's only one test, I'd simply use `test()`

src/components/MaText/MaText.vue
```diff
@@ -0,0 +1,85 @@
+<template>
+  <component
+    :is="tag"
+    :style="computedStyle"
+    :class="computedClass"
+    :italic="italic"
```
afontcu, line 6:
```suggestion
```

src/components/MaText/MaText.vue
```diff
@@ -0,0 +1,85 @@
...
+      }
+    },
+  },
+}
+</script>
+
+<style scoped>
+.ma-text {
+  padding: 0;
+  margin: 0;
```
afontcu, line 79: neat!

src/components/MaText/MaText.spec.js
```diff
@@ -0,0 +1,64 @@
...
+      expect(getByText(slotText)).toHaveStyle(`color: var(--color-${color})`)
+    })
+  })
+})
+
+const renderComponent = (props = {}) => {
+  return render(
+    MaText,
+    {
+      props: { ...props },
```
afontcu, line 57: pointless, stylistic suggestion 💡:

```suggestion
      props,
```

src/components/MaText/MaText.stories.js
```diff
@@ -0,0 +1,55 @@
...
+            :key="textSize"
+            :size="textSize"
+            :tag="tag"
+            :italic="italic"
+            :bold="bold"
+            :tone="tone"
+          >
+            {{ textSize }} example text
+          </ma-text>
+        </ma-stack space="small">
```
afontcu, line 31:
```suggestion
        </ma-stack>
```

src/tokens/tones.js
```diff
@@ -2,12 +2,12 @@ const color = require('./color')
...
-  neutral: color.gray.dark,
-  muted: color.gray.base,
+  red: color.red.base,
+  pink: color.pink.base,
+  blue: color.blue.base,
+  green: color.green.base,
+  yellow: color.yellow.base,
+  'gray-darker': color.gray.darker,
+  'gray-dark': color.gray.dark,
+  'gray-base': color.gray.base,
```
afontcu, line 12: wondering if this one could just be `gray`?

edufarre: I believe we should decide which naming we would like to use for the tones and then go for it 😞

afontcu: we kinda did, right? the thing is that other `-base` colors do not use the sufix

edufarre: You're right, gonna change it then 😄

edufarre: Done! See the related [commit](https://github.com/holaluz/margarita/pull/327/commits/64bf28496f32d0b5a6729edc14e6c06499123292)

alexcastelo95 (approved):
I think I understand all the code. Apparently, I'm only left with the question of what convention we use with the colors and shades. 🤔 

Other than that, I think it's a very good job. 🚀

src/tokens/text.js
```diff
@@ -0,0 +1,39 @@
...
+    medium: {
+      'font-size': '1.25rem',
+      'line-height': 1.35,
+    },
+  },
+}
+
+module.exports = {
+  textSize,
+}
```
Samitier, line 39: line height is the same all the time. Why is it here?

afontcu: thought about it, too, but I believe it is good to make it explicit?

src/components/MaText/MaText.stories.js
afontcu: I think the conversation was resolved but the code wasn't changed? 😂

src/components/MaText/MaText.vue
```diff
@@ -0,0 +1,105 @@
...
```
Samitier, line 94: we are still not correcting the line-height offset in the first and last line of text. If we don't do that, the spacing tokens won't work correctly because they will change with font-size.

In order to have a correct separation between text elements, the ma-text component should have negative top and bottom margins proportional to the font-size. That way the text will "start" at the beginning of the text.

Example without the margin (the image and text are not aligned due to the line-height in the title):
image


Example with the negative margin:
image


If I'm not mistaken, if we have a line-height of `1.35`, we should have a `margin: -.35em 0`. `em` instead of `rem` because we want it relative to the font-size of the ma-text.

src/components/MaText/MaText.vue
```diff
@@ -0,0 +1,105 @@
...
+      default: false,
+    },
+
+    /**
+     * Applies bold weigth to text
+     */
+    bold: {
+      type: Boolean,
+      default: false,
+    },
```
Samitier, line 53: since this component has a slot we will have weird behaviours with this props. Example:
```vue
<ma-text :bold="false" :italic="false">
 <b>What should this render?</b> <i>I dunno</i>
</ma-text>
```

Why don't we stick to `colors` and `sizes` only?

afontcu: I'd say… there's no need to use `b` or `i` if we have MaText :)

Samitier: then maybe the content should be a string prop instead of a slot?

src/tokens/text.js
edufarre: M.. as far as I know, and I looked at how tailwind does handle this, `font tokens` represent the `line-height` and the `font-size` of the text. This being said, and if you think that it's enough, I would something like:
```js
// tokens/text.js
const lineHeight = 1.35
...
```
and then use it when building our tokens, but, I would rather have it there than just set it as default on the component.

src/components/MaText/MaText.vue
edufarre: Yep, you're right! But, I believe we decided to tackle the `negative margin` issue in the next iteration, that's why I did not implement it.

src/components/MaText/MaText.vue
edufarre: If we use a string, what about doing a composition of `MaText` components? 🤔 
i.e:
```html
<ma-text>Hello <ma-text tone="blue"> how are you</ma-text></ma-text>
```

afontcu: Also, and a bit unrelated to everything else exposed here, I feel the slot API is more _friendly_ to newcomers than the prop one, because it "just looks" like an HTML element. This is highly opinionated, of course xD

src/components/MaText/MaText.vue
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * This is our custom text component wich follows our design system guidelines
+ */
+export default {
+  name: 'MaText',
+
+  props: {
+    /**
+     * Sets the text element tag
+     */
```
afontcu, line 23:
```suggestion
    /**
     * Set the HTML element tag to ensure the document semantics are meaningful
     */
```

src/components/MaText/MaText.vue
```diff
@@ -0,0 +1,105 @@
...
+  >
+    <slot />
+  </component>
+</template>
+
+<script>
+import { text, tones } from '../../tokens'
+/**
+ * This is our custom text component wich follows our design system guidelines
+ */
```
afontcu, line 16:
```suggestion
/**
 * Renders text following the Design System guidelines
 */
```

afontcu (approved):
Just suggested some small tweaks on descriptions, but it looks solid overall 👍 as mentioned, we can keep on iterating the component!

src/components/MaText/MaText.vue
Samitier: I think the component is a bit useless without it no? It breaks our own design system 😂 

And if we add it later it will be a breaking change. Why don't we add it now? It's one line of css.

afontcu: Hm, it is true that if line-height is stable and we have a single font family, fixing the offset is kinda trivial

afontcu: (bear in mind we should handle both the "top" offset and the "bottom" one, probably by using `::before` and `::after`)

afontcu: https://css-tricks.com/how-to-tame-line-height-in-css/

afontcu: we should also take into account cap letters with accents: `À`

edufarre: Solved! Thanks for the support @afontcu .

Pinging @Samitier :
Screen Shot 2021-01-28 at 17 29 30

PR #328, opened 2021-01-18
fix: Fix dashed names on tokens

afontcu:
...by creating a custom style dictionary transformer.

This PR is actually a breaking change but I'm quite sure noone has used CSS custom props yet, so let's keep it simple a release a patch version so everyone has the right values available :) Correct me if I'm wrong!

Before:

```css
  --spacing-2-x-large: 4rem;
  --spacing-3-x-large: 5rem;
  --spacing-4-x-large: 6rem;
  --spacing-5-x-large: 8rem;
  --spacing-6-x-large: 11rem;
```

After:

```css
  --spacing-2x-large: 4rem;
  --spacing-3x-large: 5rem;
  --spacing-4x-large: 6rem;
  --spacing-5x-large: 8rem;
  --spacing-6x-large: 11rem;
```

Same thing with SCSS variables.

---

btw, if you want to discuss naming and have `2xlarge` available instead of `2x-large`… now's the time 😇


style-dictionary.config.js
```diff
@@ -1,8 +1,23 @@
...
+  type: 'name',
+  transformer: (prop) => prop.path.join('-'),
+})
+
+StyleDictionary.extend({
   source: ['.style-dictionary/index.js'],
   platforms: {
     css: {
-      transformGroup: 'css',
+      transforms: [
```
afontcu, line 13: I took transforms from the docs:

https://amzn.github.io/style-dictionary/#/transform_groups?id=scss
https://amzn.github.io/style-dictionary/#/transform_groups?id=css

edufarre (approved):
Couldn't be happier to see a PR like this! Love that we finally decided using more potential that `style dictionary` offers 🚀 

Just a quick question, what about changing the `tokens:build` of the package.json? style-dictionary.config.js is not a 'plain' config file anymore, shall we just trigger the written script with `node`? Just saying this because we currently trigger it passing the file like it was just config, and when executing the code, the build is already happening when we do `.buildAllPlatforms()`

edufarre:
⚠️ Letter to the kings: And yeah! I would go for `2x-large` instead of `2-x-large` ⚠️ 
(pls don't kill me, otherwise, the kings won't find to whom should they reply) 🤣

afontcu:
> Just a quick question, what about changing the `tokens:build` of the package.json? style-dictionary.config.js is not a 'plain' config file anymore, shall we just trigger the written script with `node`?

Great idea, done! Now I could wrap everything style-dictionary-related together 🎉

edufarre (approved):
Wicked!

Samitier (approved):

PR #329, opened 2021-01-25
fix: Replace .scss component styles with .css

afontcu:
Let's see how thing look like after removing SCSS from some components and replace it with the canonical postcss implementation (`postcss-preset-env`).

---

Okay, let me share some stuff:

* As we mentioned, postcss was already installed so adding a new plugin was quite simple.
* I leveraged postcss-preset-env, which is the recommended way to use future CSS features, today – even in [the homepage of postcss](https://postcss.org/).
* Stage 2 is the default preset value, I made it explicit. Feels like a sensible default, but open for debate!
* Please notice that even though I added postcss-nesting, **it does not behave exactly the same as SCSS nesting**. Some things needed to change, and some fancy stuff such as `&__whatever {}` no longer works.
* Git diff looks enormous because git is unable to tell that `.scss` files were renamed to `.css` AND heavily modified (mostly due to indentation). I tried with `git mv` with no luck.
* I did not try to fix legacy CSS usage. There are several components out there that could use some revamping, but this was out of scope for this PR.

I'd appreciate a thorough review! I'm not entirely sure we could/should merge this without a release, because a lot could go wrong.


src/scss/_sass-mq.scss
```diff
@@ -11,4 +11,4 @@ $mq-breakpoints: (
 $mq-and-landscape: '(orientation: landscape)';
 $mq-and-portrait: '(orientation: portrait)';
 
-@import '~sass-mq';
+@import 'node_modules/sass-mq/mq';
```
afontcu, line 14: Not entirely sure why webpack's `~` alias stopped working

postcss.config.js
```diff
@@ -1,5 +1,11 @@
 module.exports = {
   plugins: {
     autoprefixer: {},
+    'postcss-preset-env': {
+      stage: 2,
+      features: {
+        'nesting-rules': true,
+      },
+    },
```
edufarre, line 9: Question: Will this replace the current `stylelint` functionalities we have set across our different projects?
See: https://github.com/holaluz/npm-scripts/blob/master/stylelint.config.js

I've never used `postcss` before, so... pardon me if the question seems stupid 😅

afontcu: hmm they are not related – stylelint is a linter, while [postcss](https://postcss.org/) is a preprocessor

src/components/MaAlert/MaAlert.vue
```diff
@@ -49,4 +49,4 @@ export default {
 }
 </script>
 
-<style lang="scss" src="./MaAlert.scss"></style>
+<style src="./MaAlert.css"></style>
```
Samitier, line 52: scoped? 🙏 👼 

but it might be a breaking change if we use that class somewhere

src/components/MaAlert/MaAlert.css
```diff
@@ -0,0 +1,82 @@
...
+
+.alert-banner--large {
+  padding: 0.75rem;
+}
+
+.alert-banner--error {
+  --alert-color: var(--color-red-base);
+  --alert-border-color: var(--color-red-base);
+  --alert-background-color: var(--color-red-light);
+  --alert-icon: url('../../assets/icons/alert-error.svg');
```
Samitier, line 60: just out of curiosity: is the assets alias working? I don't know if it has ever worked in margarita or just in the nuxt projects:
`url('~assets/icons/alert-error.svg)` or `url('@/assets/icons/alert-error.svg)`

src/components/MaAlert/MaAlert.css
```diff
@@ -0,0 +1,82 @@
...
+  --alert-background-color: var(--color-green-light);
+  --alert-icon: url('../../assets/icons/rounded-tick.svg');
+}
+
+.alert-banner--warning {
+  --alert-color: var(--color-yellow-dark);
+  --alert-border-color: var(--color-yellow-dark);
+  --alert-background-color: var(--color-yellow-light);
+  --alert-icon: url('../../assets/icons/warning.svg');
+}
```
Samitier, line 82: where are the media queries? Or is still wip?

src/components/MaAlert/MaAlert.vue
afontcu: hahahaa followed the same thought process. Was about to set the `scoped` attribute when I thought about custom modifications… which should not happen, to be honest 😬

src/components/MaAlert/MaAlert.css
afontcu: still WIP, even though I'm not quite sure the responsive behavior is expected in the new Alert design 🤔 I'll ping the design crew and see what we come up with!

afontcu: checked with @icurdu. The responsive behavior of Alert's padding was not expected ---> 🔥 

😃

src/components/MaAlert/MaAlert.css
afontcu: Don't really know 😂  AFAIK aliases are webpack's territory, so I'd expect the same behavior. That being said, it might depend on which loader gets loaded first

src/components/MaAlert/MaAlert.vue
Samitier: hahah maybe we can consider doing this for the next breaking change. There are more components in margarita that are unscoped too. I vote for being consistent and either scope them all or unscope all.

src/components/MaButton/MaButton.css
```diff
@@ -0,0 +1,159 @@
...
+}
+
+.ma-button--primary {
+  background-color: var(--color-pink-base);
+  color: var(--color-white-base);
+
+  &:focus {
+    outline: 2px solid var(--color-pink-light);
+    outline-offset: 0;
+  }
```
Samitier, line 57: this 2 lines of css are repeated 4 times in this document. What happens if we add them only in `.ma-button`?

src/components/MaCard/MaCard.css
```diff
@@ -0,0 +1,16 @@
...
+.ma-card--gray {
+  background-color: var(--color-brown-light);
+}
+
+.ma-card--white {
+  box-shadow: 0 0 6px 2px var(--shadow-light);
+  background-color: var(--color-white-base);
+}
+
+.ma-card--has-padding-top {
```
Samitier, line 14: 😂

Samitier (approved):
amazing job! 🔝

src/components/MaCard/MaCard.css
afontcu: kill me 😂

src/components/MaButton/MaButton.css
afontcu: fixed! (that being said, I think refactoring components' CSS should be done when we focus on updating them – as I mentioned, I didn't focus on improving current CSS, but rather removing SCSS for good 😉)

edufarre (approved):
Good job!

⚠️ Let's bear in mind if we should add `scoped` styles to all the components or maybe remove it from the ones that they currently have it, but, whatever we decide, let's have consistency across all of them 💯

afontcu:
OK, this looks solid (or as solid as it could look like).

I think I'll issue a patch version because we removed the responsive behavior of Alerts, which were not expected. This will also allow us to test all these changes in the wild – nothing should break, but changes are substantial and the sooner we test them, the better.

PR #332, opened 2021-01-25
fix(perf): Remove normalization for older browsers

afontcu:
We're also generating CSS from CSS files… instead of using SCSS! 🎉 

---

* Import normalize from a postcss plugin, so that it uses browserslist config.
* Load Ciutadella manually (so we avoid using SCSS mixin). Will work for now.
* Move all partials form `/scss` to `/css`.

As far as I can tell, margarita's output remains the same (because we import the new `index.css`). CSS output looks exactly the same (a bit shorter due to postcss-normalize behavior – it uses browserslist to exclude some useless rules).


src/css/partials/ciutadella.css
```diff
@@ -0,0 +1,79 @@
...
+      format('truetype'),
+    url('https://assets.holaluz.com/fonts/ciutadella/Ci1001-Md.svg#str-replace(%22Ciutadella%22,%20%22%20%22,%20%22_%22)')
+      format('svg');
+}
+
+@font-face {
+  font-family: Ciutadella;
+  font-weight: 700;
+  font-display: swap;
+  src: url(https://assets.holaluz.com/fonts/ciutadella/Ci1001-SmBd.eot)
```
Samitier, line 53: maybe nitpicking, but could we use local() to save some data for those who have that font downloaded already?

src/css/partials/visually-hidden.css
```diff
@@ -1,4 +1,4 @@
-// https://snook.ca/archives/html_and_css/hiding-content-for-accessibility
+/* https://snook.ca/archives/html_and_css/hiding-content-for-accessibility */
 .visually-hidden {
```
Samitier, line 2: what is this class?

the article linked is from 2011, is this still relevant?

src/css/index.css
```diff
@@ -0,0 +1,17 @@
+/**
+ * Auto-generated CSS custom properties from style-dictionary
+ */
+@import 'tokens.css';
+
+/**
+ * Make browsers render all elements more consistently
+ * https://necolas.github.io/normalize.css/
+ */
+@import 'normalize.css';
```
Samitier, line 10: maybe we could start considering if normalize is really necessary given our supported browsers (and given that the project is dead since 2018). Maybe with a:
```css
* {
   box-sizing: border-box;
   margin: 0;
   padding: 0;
}
```
we have more than enough, no?

src/css/partials/visually-hidden.css
afontcu: yep! it is actually used in several components – for instance, MaOption. We (visually) hide the real input and style a different one to make it look as we please, but keeping everything a11y-friendly

src/css/index.css
afontcu: yeah, that's true… we might want to check it out first, but getting rid of normalize is definitely something worth exploring

afontcu: we could even consider sanitize.css: https://github.com/csstools/sanitize.css

edufarre: I believe you both are right! Shall we resolve this comment and open an `issue` thus iterating with the other peers what should we do or which approach suits us better?

afontcu: makes total sense 👌  on it

edufarre (approved):
🚀

PR #337, opened 2021-01-28
chore: Add UI regression tests

edufarre:
Yesterday while fixing some styling issues with @afontcu and @alexcastelo95, we agreed on having visual tests in order to avoid having any styling issues and releasing a new version of the project anyway without being aware of them. We first thought about [Percy.io](https://percy.io/), but then we bumped into a library (_Made by Storybook maintainers_) that could help us deal with these situations on top of having the entire Design System Team (Design + Front-end) on the loop.

The library is called [Chromatic](https://www.chromatic.com/) and as you will see, how easy is to add it to the project.

This PR aims to show how we could test our components UI from now on. At the same time, Chromatic a preview version of our Storybook, which is kinda big `win-win` situation, isn't it? 🚀 

Bear in mind that due to this is the first PR using it, we still need `baseline snapshots` which will be the reference and compared with the different PR targeting the `master` branch. What I mean by that is that we won't have the entire Chromatic functionalities in this proof of concept.

You will see that Chromatic adds a few more check on each PR:
Screen Shot 2021-01-28 at 08 53 28
And if you are subscribed to this project using the `Github` app on Slack, you will see it there as well:
Screen Shot 2021-01-28 at 08 53 38

Moreover, we could always integrate the [Chromatic notifications on Slack](https://www.chromatic.com/docs/integrations#slack-notifications) in our `#design-system` channel, in order to always have tracked the status of our visual regression tests. I would like to add, that in the Power Rangers team, we do have Percy notifications in our alert channel, thus making everyone in the team aware of what it does change in each PR (and then Iban can always review the snapshots and let comments or approve the changes)


alexcastelo95 (approved):
Sounds exciting.  🚀

.github/workflows/ci.yml
```diff
@@ -31,5 +33,14 @@ jobs:
       - name: Run tests
         run: npm run test
 
+      - name: Run UI regression tests
```
afontcu, line 36: It is great that we integrate this in our CI pipeline workflow because we save github actions quota: we only install deps once, and we don't run a step if the previous one has failed 🎉

afontcu (approved):
👌

afontcu:
btw, if this get merged I'll remove the Netlify integration (currently paused), as it isn't needed – chromatic+storybook already provides a PR preview URL 🎉

afontcu:
btw, it is no longer a `[PoC]`, is it? 😉 It shouldn't trigger a release either, so I'd tag it as a `chore: ` or similar

edufarre:
> btw, it is no longer a `[PoC]`, is it? 😉 It shouldn't trigger a release either, so I'd tag it as a `chore: ` or similar

In my opinion it still is a PoC, dunno if the rest of the team is okay with this new step in our CI pipeline tho. But yeah, you're, this should be a `chore` release

afontcu:
> > btw, it is no longer a `[PoC]`, is it? 😉 It shouldn't trigger a release either, so I'd tag it as a `chore: ` or similar
> 
> In my opinion, it still is a PoC, dunno if the rest of the team is okay with this new step in our CI pipeline tho. But yeah, you're, this should be a `chore` release

right! 👍👍👍 let's just make sure we create the appropriate commit message 😉

afontcu (approved):

PR #339, opened 2021-01-28
[PoC] Functional Stack

afontcu:
By making Stack functional we can successfully style slot's content while maintaining scoped styles.

The downsides of the current implementation is that we cannot set additional attributes to a Stack (other than its props), because it will swallow them instead of rendering/passing them down to its root child. See the Modal changes.

### DIFF
--- a/src/components/MaModal/MaModal.vue
+++ b/src/components/MaModal/MaModal.vue
@@ -11,32 +11,33 @@
             data-testid="overlay"
             @click="closeModal"
           />
-          <ma-stack
-            ref="modal"
-            :aria-label="title"
-            :class="`modal--width-${width}`"
-            space="medium"
-            role="dialog"
-            aria-modal="true"
-            class="modal"
-          >
+          <ma-stack space="medium">
             <div
-              :class="`modal-header--${headerType}`"
-              class="modal-header"
-              data-testid="modal-header"
+              ref="modal"
+              :aria-label="title"
+              :class="`modal--width-${width}`"
+              role="dialog"
+              aria-modal="true"
+              class="modal"
             >
-              <span class="modal-title">{{ title }}</span>
-              <ma-button
-                category="no-background"
-                data-testid="close-button"
-                class="icon-close"
-                @click="closeModal"
+              <div
+                :class="`modal-header--${headerType}`"
+                class="modal-header"
+                data-testid="modal-header"
               >
-                <ma-icon icon="Close"> </ma-icon>
-              </ma-button>
-            </div>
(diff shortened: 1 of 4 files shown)


src/components/MaModal/MaModal.vue
```diff
@@ -172,7 +173,7 @@ export default {
       // If we cannot find the modal let's fail gracefully.
       if (!modal) return
 
-      this.focusableElements = modal.$el.querySelectorAll(FOCUSABLE_ELEMENTS)
+      this.focusableElements = modal.querySelectorAll(FOCUSABLE_ELEMENTS)
```
afontcu, line 176: Since its ref="modal" no longer returns a component, we don't need to access `$el`

src/components/MaModal/MaModal.vue
```diff
@@ -11,32 +11,33 @@
...
-          <ma-stack
-            ref="modal"
-            :aria-label="title"
-            :class="`modal--width-${width}`"
-            space="medium"
-            role="dialog"
-            aria-modal="true"
-            class="modal"
-          >
+          <ma-stack space="medium">
```
afontcu, line 14: Only change here is that I removed all attrs from the Stack and moved them to an inner div:

```vue
<ma-stack space="medium">
  <div ref="modal"....>
```

src/components/MaStack/DemoComp.vue
```diff
@@ -0,0 +1,17 @@
...
+<script>
+export default {
+  name: 'DemoComp',
+}
+</script>
+
+<style scoped>
+p {
+  color: blue;
+}
```
afontcu, line 16: This is what doesn't work when using the current Stack implementation.

src/components/MaModal/MaModal.vue
carlotamolintr: nicee 👌👌👌

afontcu: @carlotamolintr do you recall if any of your MaStack has additional attributes? We'd need to change that, and I'd like to check that it's doable ;)

carlotamolintr: We used the ma-stack as a normal <div>. Inside this could be a button and slot.  So, iI guess we don't have additional attributes.🤔🤔🤔

Samitier:
I used this implementation of stack in https://github.com/holaluz/dry-martini/pull/420. I also moved the column component to functional too but still doesn't work with scoped styles. I guess I have to make the LayoutComponent funcional too even though it just renders stacks and cols internally 😞

afontcu:
Fixed in #376

PR #342, opened 2021-01-29
chore: Include vetur folder and design tokens back to git

### LINKED ISSUE #334: Should we add vetur/ folder and tokens back to git?
They are currently [excluded from git](https://github.com/holaluz/margarita/blob/master/.gitignore#L19-L26) because they are autogenerated. 

I believe, though, that it would make sense to be able to track changes when reviewing a PR that affects either `vetur/` or tokens.

Thoughts? @holaluz/frontend

afontcu:
Closes #334

### DIFF
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -1,3 +1,12 @@
 const lintStagedConfig = require('@holaluz/npm-scripts').lintStaged
 
-module.exports = lintStagedConfig
+module.exports = {
+  ...lintStagedConfig,
+  '*.{js,vue}': [
+    'npm run vetur:build',
+    'prettier --write',
+    'eslint',
+    'npm run test -- --findRelatedTests',
+  ],
+  './src/tokens/*.js': ['npm run tokens:build'],
+}
--- a/package.json
+++ b/package.json
@@ -16,7 +16,6 @@
     "vetur"
   ],
   "scripts": {
-    "prepare": "npm run tokens:build && npm run vetur:build",
     "build": "vue-cli-service build --target lib --name margarita src/index.js",
     "format": "prettier 'src/**/*.{js,vue,css}' --write && stylelint 'src/**/*.scss' --fix",
     "lint": "npm run lint:js && npm run lint:css",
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -16,7 +16,6 @@
     "vetur"
   ],
   "scripts": {
-    "prepare": "npm run tokens:build && npm run vetur:build",
```
afontcu, line 19: If I'm not mistaken we don't need to prebuild these files anymore

package.json
```diff
@@ -27,7 +26,7 @@
     "storybook:serve": "vue-cli-service storybook:serve -p 6006 -c .storybook",
     "test": "vue-cli-service test:unit",
     "test:watch": "vue-cli-service test:unit --watch",
-    "tokens:build": "node .style-dictionary/build.js && prettier 'src/scss/*' --write",
+    "tokens:build": "node .style-dictionary/build.js",
```
afontcu, line 29: Precommit hook will handle prettier for us

lint-staged.config.js
```diff
@@ -1,3 +1,12 @@
...
-module.exports = lintStagedConfig
+module.exports = {
+  ...lintStagedConfig,
+  '*.{js,vue}': [
+    'npm run vetur:build',
+    'prettier --write',
+    'eslint',
+    'npm run test -- --findRelatedTests',
+  ],
+  './src/tokens/*.js': ['npm run tokens:build'],
```
afontcu, line 11: Not 100% sure about this. I mean, not sure if this is enough to make sure vetur and tokens are properly updated when a new PR is set up. I guess it should, but…

thoughts?

edufarre: M... I am not 100% sure about this. Wouldn't it this be better to set it in our CI (and just trigger the `vetur` and `tokens` build when `pushing to master` or on `pull request` ?

package.json
edufarre: Not sure about the `tokens:build` tho 🤔  I do believe that we set that in order to avoid running the `build` command on our CI pipelines before packaging our project. I might be wrong tho.

Screen Shot 2021-01-29 at 13 10 34

afontcu: hm, but we don't need to add those tokens to the repo if they are already committed, right?

lint-staged.config.js
afontcu: then we'd need to commit those files back to the PR branch, right? we'd need to set a different workflow (not CI) and use something like this: https://github.com/stefanzweifel/git-auto-commit-action

afontcu: hm, also – if we add tokens only on push to master, we end up with the same situation: you're not aware about the changes on tokens/vetur on PR review time

edufarre: You're right. Sorry I missed that out. Actually, I believe that if we do it I our CI would require too much boilerplate (like using the `auto-commit-action` that you previously commented)

edufarre (approved):

afontcu:
Let's merge it as is for now, and see if the workflow makes sense. We can always iterate on it.

PR #344, opened 2021-02-01
docs: Use storybook controls 

edufarre:
Replacing story knobs. Now we do have a table with the component props/API:
Screen Shot 2021-02-01 at 09 35 35
⚠️ However, beware that the `String type` properties that must be an element of a specific number of options (string[]) in order to be a valid value, do not parse to a `select` in the controls table, that why it must be set by hand. I reckon it might not be the fanciest solution, but, it does de trick.

In addition, we do have component IntelliSense as well:
Screen Shot 2021-02-01 at 09 43 14
Screen Shot 2021-02-01 at 09 41 36


afontcu (commented):
Looks promising! Hopefully both vue-docgen-api and storybook will implement the missing pieces to make everything work flawlessly when it comes to prop `@values` 🎉

src/components/MaButton/MaButton.stories.js
```diff
@@ -1,96 +1,43 @@
...
+  click: action('click'),
 }
 
-export const IconButton = () => {
-  const icon = select('Icons', availableIcons, availableIcons[0])
-  const iconSize = number('Icon height size in px', 18)
+const getProps = (argTypes) =>
+  Object.values(argTypes)
+    .filter((a) => a.table?.category === 'props')
+    .map((a) => a.name)
```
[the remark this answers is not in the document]
edufarre: Done. I did a bit clean up into the component's story. However, I reckon that the way I do handle the `click` action, might not be the best one. I did not figure out a better approach.

sergigraciaj (approved):

src/components/MaButton/MaButton.stories.js
```diff
@@ -1,96 +1,37 @@
...
     },
-  }
+    /**
+     * Tell Storybook that 'click' arg to your story should be an action
+     */
+    click: { action: 'click' },
+  },
 }
 
-export const IconButton = () => {
```
afontcu, line 66: We lost the IconButton story, right?

src/components/MaButton/MaButton.vue
```diff
@@ -29,30 +22,45 @@ export default {
...
+     * Sets the component displaying style
+     */
     category: {
       type: String,
       default: 'primary',
-      validator: (v) => AVAILABLE_CATEGORIES.includes(v),
+      validator: (v) =>
+        ['primary', 'secondary', 'white', 'gradient', 'no-background'].includes(
+          v
+        ),
```
afontcu, line 63: This change won't be needed once this PR gets merged 😉 
https://github.com/vue-styleguidist/vue-styleguidist/pull/1051

src/components/MaButton/MaButton.vue
```diff
@@ -29,30 +22,45 @@ export default {
   },
 
   props: {
+    /**
+     * The tag property of the component sets when the component tag
```
afontcu, line 26:
```suggestion
     * Sets the HTML tag element of the component
```

cillascillas (approved):
It looks great in the preview! Is there any way of hiding the knobs tab now that it's empty (or any other that we don't use)?

src/components/MaButton/MaButton.stories.js
edufarre: Yep, I do believe this was on the scope of this PR. However, as far as I know, we should implement/refactor the `MaIcon` store and then use that one inside our `MaButton.stories`, thus we won't need to maintain more code than the required one
See: https://storybook.js.org/docs/vue/workflows/stories-for-multiple-components#reusing-subcomponent-stories

.storybook/main.js
```diff
@@ -8,5 +8,6 @@ module.exports = {
     '@storybook/addon-knobs',
     '@storybook/addon-links',
     '@storybook/addon-viewport',
+    '@storybook/addon-essentials',
```
afontcu, line 11: If we add the essentials addon we can then remove [the ones included in the package](https://storybook.js.org/docs/react/essentials/introduction), right?

edufarre:
> It looks great in the preview! Is there any way of hiding the knobs tab now that it's empty (or any other that we don't use)?

I do believe we should remove the `@storybook/addon-knobs` addon

.storybook/main.js
edufarre: Yeah, I forgot about it. My bad.

cillascillas: I'll keep an eye on that so I can remove `addons-docs` from the Docs PR!

edufarre:
@cillascillas and @afontcu after checking the project dependencies, I noticed that we cannot delete `@storybook/addon-actions` as we still use it in other stories (we should keep it there until we refactor ALL the remaining component stories)

I deleted:
-  `@storybook/addon-links`
-  `@storybook/addon-viewport`

src/components/MaButton/MaButton.stories.js
afontcu: Not sure I 100% follow you here – what I mean is that `IconButton` story is gonne (if I'm not mistaken!) and I think is quite a common scenario that we might want to maintain documented (so we can use it as a diff tool when we improve either the Button or the Icon component).

PR #351, opened 2021-02-05
chore: Abstract render function for tests

afontcu:
So we don't need to initialize Margarita env (with responsible plugin and so on) every time.

### DIFF
--- a/src/margarita-test-utils.js
+++ b/src/margarita-test-utils.js
@@ -1,0 +1,35 @@
+/** @typedef {typeof import('@testing-library/vue').render} Render */
+
+import { render as VTLrender } from '@testing-library/vue'
+import responsivePlugin from './plugins/responsivePlugin'
+
+/** @type {Render} */
+function render(component, options = {}, callbackFn = () => {}) {
+  return VTLrender(
+    component,
+    {
+      ...options,
+      stubs: {
+        transition: { template: '<div><slot /></div>' },
+      },
+    },
+    (vue) => {
+      // this.$layout.currentBreakpoint will be 'mobile' by default
+      vue.use(responsivePlugin)
+
+      callbackFn(vue)
+    }
+  )
+}
+
+// Export useful packages from Vue Testing Library
+export {
+  fireEvent,
+  logDOM,
+  logRoles,
+  waitFor,
+  waitForElementToBeRemoved,
+} from '@testing-library/vue'
+
+// and override the render method with our own implementation
+export { render }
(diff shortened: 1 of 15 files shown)


src/margarita-test-utils.js
```diff
@@ -0,0 +1,26 @@
+/** @typedef {typeof import('@testing-library/vue').render} Render */
```
edufarre, line 1: Smart!

src/margarita-test-utils.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+// Export useful packages from Vue Testing Library
+export {
+  fireEvent,
+  logDOM,
+  logRoles,
+  waitFor,
+  waitForElementToBeRemoved,
+} from '@testing-library/vue'
```
edufarre, line 23: Are you exporting this in order to make more emphasis on them instead of exporting everything from `@testing-library/vue` plus the custom render function (as we do in the other projects)?

Just curios about it 😄

edufarre (approved):

src/margarita-test-utils.js
afontcu: yeah, that's mostly it 👍  Also, so that when autocompleting `getByText` (for example) in our tests VSC doesn't get tricked and import it from the package instead of from `render()`.

edufarre:
Other small question. Would import the render function doing something like `import { render } from 'margarita-test-utils'` do the trick instead of `import { render } from '@margarita/margarita-test-utils'`?

src/margarita-test-utils.js
edufarre: Noice! Thanks

afontcu:
> Would importing the render function doing something like `import { render } from 'margarita-test-utils'` do the trick instead of `import { render } from '@margarita/margarita-test-utils'`?

Not really, because the former only works due to [our Jest config](https://github.com/holaluz/npm-scripts/blob/master/jest.config.js#L14), while the latter is understood by both Jest and VSC, and thus we can get types. Otherwise VSC isn't able to get types from file (as far as I could tell)

sergigraciaj:
I understand what you do in this PR, but I don't get the point of this PR, could you explain the purpose of this PR a bit more, please?

afontcu:
> I understand what you do in this PR, but I don't get the point of this PR, could you explain the purpose of this PR a bit more, please?

It is mostly to (I quote) "So we don't need to initialize Margarita environment (with the responsible plugin and so on) every time.". For instance, currently we need to install the responsive plugin for every component that features a responsive prop (https://github.com/holaluz/margarita/pull/351/files#diff-bae0f47e9b4244822c1c56a4a81ea12491c6522f6f900080a589c51acc8bee50L59-L61).

It is the same reasoning behind creating the `XXXX-test-utils` we are using on other apps

sergigraciaj (approved):
Thank you!

PR #353, opened 2021-02-09
fix: Replace MaAlert title and text components and add Controls

cillascillas:
- [x] Add MaHeading and MaText as title and text components
- [x] Replace knobs with controls in docs
- [x] Review MaAlert tests


src/components/MaAlert/MaAlert.vue
```diff
@@ -1,39 +1,58 @@
...
+  },
+
   props: {
+    /**
+     * Sets the size of the component
+     */
     size: {
       default: 'medium',
       type: String,
       validator: (value) => AVAILABLE_SIZES.includes(value),
```
afontcu, line 37: For now, vue-docgen-api (and thus Vetur) are unable to get the right values from a variable, so they can't provide them as autocomplete options.

However, this works:

```suggestion
      validator: (value) => ['small', 'medium', 'large'].includes(value),
```


It is a known limitation and [we're working on it](https://github.com/vue-styleguidist/vue-styleguidist/pull/1051) :)

edufarre: I do believe we could do the same one the `type` property.
```js
// from:
validator: (value) => AVAILABLE_TYPES.includes(value),

// to:
validator: (value) => ['error', 'info', 'success', 'warning'].includes(value),
```

afontcu:
btw, even if a refactor, I do believe this PR "deserves" being released, as it kind of fixes the Alert UI. I'd tag it as a `fix: `  😉

cillascillas:
We've been discussing that with @edufarre before and we decided `refactor:` could do it. However, if we tag it as `fix:`, as you suggested, it will be "available" sooner, so it's ok for me too!

src/components/MaAlert/MaAlert.vue
```diff
@@ -1,43 +1,61 @@
 <template>
   <div :class="['alert-banner', getClasses]">
     <span class="alert-banner__icon" />
-    <div class="alert-banner__content">
-      <p v-if="title" class="alert-banner__title" v-text="title" />
+    <ma-stack space="xsmall">
+      <ma-heading v-if="title" size="xsmall" v-text="title" />
+      <!-- @slot Alert content slot -->
```
afontcu, line 6: Would it make sense to document that slot's content is only displayed if `text` prop is not present?

cillascillas: Or, what about not nesting them? Making them independent?

afontcu: yeah, but I think that this would yield a breaking change 😭 the fact that we have both a `text` prop and a slot is weird to begin with, but now we're tied to this API 🙃

edufarre: I am with Cillas here. I do believe we could make them independent. Despite changing this would be a breaking change, I do believe we do not have 500 alerts per projects, and it won't be super tough to reckon the different things that might have changed.

+1 to make them independent (I would keep they logic that if there is no text prop, do not display the element tho)

afontcu: Sure, the change might be simple, but a breaking change – an incompatible API change that might need changes on consumers. 

I'm OK if we think we need this change to yield a better API, but bear in mind we'd need to release this as a major version.

That being said (and since we've opened up the debate here) do we really need a slot and a prop? Wouldn't a slot be enough?

cillascillas: I've doing some digging on how we use the alert in the repositories where margarita is installed and:
- **Bonasera**: 50 Alerts using 4 slots, 40 texts and 11 titles
- **Kalimotxo**: 1 Alert using 1 slot
- **Maitai**: 8 Alerts using 3 slots, 4 texts and 5 titles
- **Dry-martini**: 6 Alerts using 1 slot, 4 texts and 5 titles

At the moment, `text` appears to be the most popular prop 😭   
But still, I was wondering if the API change might have such an impact on consumers. What would be the actual change to be done in all of them? Isn't it only a conflict for the cases where `text` and `slot` are both defined?

afontcu: Given the fact that `text` wraps text in a `MaText` while slot does not, what's the desired "merger" of both options? We have situations where alert content is more complex than a text, but I assume we might want to automatically wrapper "simple" use cases with `MaText` to remove the burden from the consumer.

cillascillas: After the exchange yesterday in the meeting, about aligning the content when `title` is not defined and what @afontcu just said about "controlling" at least the behaviour of `text` (considering it's the chosen use for most of the cases), let's keep the API as it is now, there's always room/space for a breaking change if it proves to be outdated in the future!

afontcu (approved):

cillascillas:
Latest update after discussing with @icurdu icon alignment: 
- The icon size has eventually been set to 1rem. (with no further alignment modifications to be taken)
- [Fixing alignment for slots](https://601199d127273f0021b037b0-zhcuaktnyj.chromatic.com/?path=/docs/plugins-markdown--v-markdown) won't be tackled now

Both issues will be revisited when the new icon set is defined.

src/components/MaAlert/MaAlert.vue
```diff
@@ -1,43 +1,61 @@
...
+    </ma-stack>
   </div>
 </template>
 
 <script>
-const AVAILABLE_SIZES = ['small', 'medium', 'large']
-const AVAILABLE_TYPES = ['error', 'info', 'success', 'warning']
+import MaStack from '@margarita/components/MaStack'
+import MaHeading from '@margarita/components/MaHeading'
+import MaText from '@margarita/components/MaText'
```
edufarre, line 18:
```suggestion
/**
 * Renders an aler component following the Design System guidelines
 *
 * [Component's API documentation](https://holaluz.github.io/margarita/?path=/story/components-alert--alert)
 */
```

edufarre (approved):

src/components/MaAlert/MaAlert.vue
cillascillas: Done!

PR #354, opened 2021-02-09
feat: add css vars for gradients

sergigraciaj:


afontcu:
>  ⚠️ The ma-range gradient is backwards: discuss if there is freedom for the gradient orientation in specifics parts or if the orientation is fixed.

Discussed this with @icurdu a few days ago and Range gradient orientation is fine, but the DS variable should still be left-right. However he can confirm this 😉

docs/tokens/gradients/gradients.stories.js
```diff
@@ -0,0 +1,53 @@
...
+  parameters: {
+    docs: { page: docs },
+  },
+}
+
+export const Gradients = () => ({
+  template: `
+    <ma-stack space="large">
+      <div v-for="(gradient, name) in gradients" :key="name">
+        <h2 v-text="getName(name)"/>
```
afontcu, line 15: MaHeading? 😬

src/components/MaButton/MaButton.css
```diff
@@ -118,14 +118,7 @@
...
 .ma-button--gradient {
-  background-image: linear-gradient(
-    to right,
-    var(--color-pink-base) 0%,
-    var(--color-orange-base) 40%,
-    var(--color-yellow-dark) 50%,
-    var(--color-orange-base) 60%,
-    var(--color-pink-base) 100%
-  );
+  background-image: var(--gradient-pink-orange-yellow);
```
afontcu, line 121: Issue here is that Button features a double-sized background, so that it can "slide" when hovering. I'm not quite sure we can handle this otherwise, so I would keep it as an standalone gradient for the button 🤷‍♀️  Let's assume a button is visible enough so that if we change all gradients, we'll notice 😂

edufarre (commented):
Overall, looks good to me. Shall we discuss what should we do with the button before shipping this to prod?

sergigraciaj:
https://github.com/holaluz/margarita/blob/66936813fb0958c48d0e2144ab6d0f26b8e18777/src/tokens/gradient.js#L4-L5

I decided to name it by its DS name, and to add just the colors and percentage, so the css calling this is responsible of the orientation (as we do in ma-range). This solves the left-to-right gradients and let us do top-to-bottom gradients if needed.

docs/tokens/gradients/gradients.stories.js
```diff
@@ -0,0 +1,54 @@
...
+    </ma-stack>
+  `,
+
+  data() {
+    return {
+      gradients: gradient,
+      stackStyle: {
+        fontWeight: 'normal',
+        fontSize: '0.75rem',
+        marginTop: '1rem',
```
afontcu, line 31: Not really important, but: any margin-top declaration smells like a missing Stack 😬

afontcu (changes requested):
Implementation looks good, but it seems like tokens… have not been generated? There are no tracked changes for `src/css/tokens.css` or its SCSS counterpart.

Didn't the precommit hook trigger the `tokens:build` script?

docs/tokens/gradients/gradients.stories.js
afontcu: (btw you could've simplify it a bit by simply using Stack as loop parent: `<ma-stack v-for="...">`

edufarre (approved):
Shall we update the `story` in order to be an written in `mdx` thus just having **1 file** to maintain for its `story` and `docs`>

However, we could tackle this in a future iteration

edufarre (approved):
Good job mate! 🚀  

Do not forget to update the branch with the latest code additions located in the master branch tho

afontcu (approved):
LGTM

I'd only fix the indentation on `docs/tokens/colors.stories.mdx` (looks like it's using a 4 spaces indentation)

sergigraciaj:
> I'd only fix the indentation on `docs/tokens/colors.stories.mdx` (looks like it's using a 4 spaces indentation)

The right one is 2 spaces indentation, yup?
I don't know why I got this issues just in this file... : /

PR #355, opened 2021-02-10
docs: Add Spacing tokens doc

edufarre:
Documentation page link -> https://601199d127273f0021b037b0-qyeivtvayb.chromatic.com/?path=/story/tokens-spacing--page

### Pros:
- Same page for 'canvas' and 'docs' page -> just maintain one file/documentation (shall we do the same with the colors and the other tokens?)

### Cons?
- Still need to figure out how to use our components in MDX (I would like to use `MaStack` here) 🆘 

### Quick win:
Stop checking on `margarita` the relation between token/prop namings and actual spacing size 💯 

Screen Shot 2021-02-10 at 11 18 46

### DIFF
--- a/docs/tokens/spacing.stories.mdx
+++ b/docs/tokens/spacing.stories.mdx
@@ -1,0 +1,23 @@
+import { Meta, Story } from '@storybook/addon-docs/blocks';
+import { spacing } from '../../src/tokens'
+import MaDatagrid from '../../src/components/MaDatagrid'
+
+<Meta title="Tokens/Spacing" />
+
+export const tokenRows = Object.entries(spacing).map(([tokenName,tokenValue])=>{
+    return {name:tokenName,value:tokenValue}
+})
+
+export const tokenColumns = [{name:'Spacing name',value:'name'},{name:'Spacing value',value:'value'}]
+
+export const Template = (args) => <MaDatagrid {...args } />
+
+# Spacing
+This is our values for spacing vertically and horizontally layout elements, based on 8px-grid. 
+
+<Story name="spacing" args={{
+     :rows: tokenRows
+     :columns: tokenColumns
+  }}>
+    {Template.bind({})}
+  </Story>
--- a/docs/tokens/spacing.stories.mdx
+++ b/docs/tokens/spacing.stories.mdx
@@ -5,7 +5,7 @@ import { spacing } from '../../src/tokens'
 
 # Spacing
 
-Consistent spacing creates visual balance, rythm and visual understanding that makes the user interface (UI) easier for our users to scan.
+Consistent spacing creates visual balance, rythm and visual understanding that makes the user interface (UI) easier to scan. All spacing is done in increments of 8 pixels, which is the basic unit of measurement for spacing.
 
 ---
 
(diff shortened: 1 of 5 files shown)


docs/tokens/spacing.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+## Layout Spacing
+
+<div style={{display: 'grid', gridAutoFlow: 'rows', gap: '2rem', marginTop: '1.5rem'}}>
+{
+  Object.entries(spacing).map(([tokenName,tokenValue]) => (
+    <div style={{display: 'grid', gridAutoFlow: 'column', gap: '3rem', maxWidth: '600px', alignItems: 'center', gridTemplateColumns: '4rem 11rem 4rem 3rem'}}>
+      <span>{tokenName}</span>
+      <div style={{width: tokenValue, height: tokenValue, backgroundColor: 'var(--color-gray-light)', justifySelf: 'center'}}/>
+      <span>{tokenValue}</span>
+      <span>{Number(tokenValue.split('rem')[0])*16}px</span>
```
afontcu, line 29: Humm, this is only true as long as user has defined a 16px as a default font-size for the browser (which is usually the case, but not always – this is the reason we use `rem` instead of `px`).

afontcu:
Question: Is it possible to hide the Canvas tab and leave the Docs only? Not sure if doable

docs/tokens/spacing.stories.mdx
edufarre: I was not 100% agree with this, but @icurdu asked for it. How could we tackle this with a different approach @afontcu ? Would love some help here

src/components/MaStack/MaStack.vue
```diff
@@ -19,6 +19,8 @@ export default {
   props: {
     /**
      * Sets the space gap between children.
+     *
+     * [Spacing tokens documentation](https://holaluz.github.io/margarita/?path=/story/tokens-spacing--page)
```
afontcu, line 23: Wow, does this display a link on VSC? 😮  It'd be amazing. I reckon JSDoc has a `@see` tag, but I guess vue-docgen-api won't parse it

edufarre: I tried `@see` and `@link`, does not look like its util. I did not see any change tho

edufarre:
> Question: Is it possible to hide the Canvas tab and leave the Docs only? Not sure if doable
AFAIK, nope 😞  However, we can just have 1 file to 'define' these stories, and that's what this PR aims as well.

See [storybook documentation](https://storybook.js.org/docs/react/writing-docs/mdx#documentation-only-mdx)

src/components/MaStack/MaStack.vue
edufarre: >Wow, does this display a link on VSC? 

You've got the screenshot in the PR description 😄

afontcu: neat 🎉🎉🎉

sergigraciaj:
> * Same page for 'canvas' and 'docs' page -> just maintain one file/documentation (shall we do the same with the colors?)

I agree with this 👍

afontcu:
> Still need to figure out how to use our components in MDX (I would like to use `MaStack` here) 🆘

I don't think this will work out of the box because storybook is built with React, so there's no way for MDX to understand Vue components. I'm not even sure adding a webpack loader would suffice… 😞

docs/tokens/spacing.stories.mdx
afontcu: I would only keep the `rem` values, and make sure we also use them in our designs (that is, do not talk about `px` in general) 💡

edufarre:
Sooooo... This has been here for a week, what shall we do with it? 🤔

afontcu:
> Sooooo... This has been here for a week, what shall we do with it? 🤔

https://github.com/holaluz/margarita/pull/355/files#r574387585

docs/tokens/spacing.stories.mdx
edufarre: Done

sergigraciaj (approved):

cillascillas (approved):

PR #356, opened 2021-02-11
fix: Add missing top placement and default z-index for Modal

afontcu:
Weirdly enough, Modal works fine on Storybook, but looks like it is not properly placed when used somewhere else ([here]( https://deploy-preview-379--maitai.netlify.app/es/)).

I simply added the same `top` value the animation was setting – but looks like it does the trick in Maitai. It displays the Modal where it belongs while also animating it.

Re z-index, we discussed this is the simplest solution for now. I've left a comment to make sure we understand this is a quick&dirty fix, and not something we should rely on ;)

Any additional test would be appreciated!

### DIFF
--- a/src/components/MaModal/MaModal.css
+++ b/src/components/MaModal/MaModal.css
@@ -10,6 +10,7 @@
 .modal {
   position: fixed;
   left: 50%;
+  top: 50%;
   transform: translate(-50%, -50%);
   width: 96vw; /* so that it doesn't reach end of viewport */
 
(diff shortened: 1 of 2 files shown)


edufarre (approved):

Samitier (approved):
yaaay 🎉

cillascillas (approved):

carlotamolintr (approved):

afontcu:
Lookin' good

image


merging and releasing!

PR #360, opened 2021-02-17
docs: Update components stories to use storybook's controls

edufarre:


src/components/MaDatagrid/MaDatagrid.vue
```diff
@@ -28,23 +32,44 @@ export default {
...
 
+    /**
+     * Sets the datagrid columns.
+     *
+     * If `column.value` not present, `column.component` is required, and viceversa.
+     * ```ts
+     *   <{
+     *    name: string,
+     *    value?: any,
+     *    component?: HTMLElement
```
afontcu, line 51: `component` is not actually an HTMLElement, but a Vue Component (it's not easy to get its type, btw)

src/components/MaHidden/MaHidden.js
```diff
@@ -1,23 +1,41 @@
...
+ *
+ * [Component's API documentation](https://holaluz.github.io/margarita/?path=/story/layout-hidden--hidden)
+ */
 export default {
   name: 'MaHidden',
 
   props: {
+    /**
+     * Hides the content below the desired breakpoint
+     * @values null, tablet, desktop
```
afontcu, line 16: Not sure if we should list `null` as a valid value – as we all might know, `null` and `undefined` values will pass any type validation if `required: true` is not set. I simply made it explicit so the validator array would read as documentation ("tablet, desktop" for `below`, and mobile, table for `above`).

I'd say we can safely remove them from both `values` and the validator.

src/components/MaHidden/MaHidden.stories.js
```diff
@@ -1,35 +1,61 @@
...
+      },
+      control: {
+        type: 'select',
+        options: ['mobile', 'tablet', null],
+      },
+    },
+    inline: {
+      type: 'boolean',
+      defaultValue: false,
+      description: 'Sets if content should be displaced inline',
```
afontcu, line 44:
```suggestion
      description: 'Sets if content wrapper should be an inline element',
```

src/components/MaIcon/MaIcon.vue
```diff
@@ -35,31 +44,49 @@ export default {
...
+    /**
+     * Sets icon's color
+     */
     iconColor: {
       type: String,
       default: 'currentColor',
     },
 
+    /**
+     * Sets icon's color
```
afontcu, line 88: title

src/components/MaOption/MaOption.stories.js
```diff
@@ -1,84 +1,67 @@
...
-    methods: {
-      onChange: action(`Change`),
-    },
   }
 }
+
+export const Checkbox = CheckboxTemplate.bind({})
+Checkbox.args = {
+  type: 'checkbox',
+}
```
afontcu, line 67: neat 👌

src/components/MaRange/MaRange.vue
```diff
@@ -40,15 +40,34 @@ import uuid from '@margarita/utils/uuid'
 
 const REQUIRED_STEP_KEYS = ['value', 'text']
 
+/**
+ * Renders a range element following the Design System guidelines
```
afontcu, line 44:
```suggestion
 * Renders a range input element following the Design System guidelines
```

src/components/MaStack/MaStack.vue
```diff
@@ -13,12 +14,25 @@ import { responsivePropValidator } from '@margarita/utils/responsivePropValidato
...
   props: {
     /**
      * Sets the space gap between children.
+     *
+     * If an array is passed, values will target the design system breakpoints.
+     * ```ts
+     * // This would apply medium on all the different breakpoints
+     * space = 'small'
+     * // This would apply none on mobile, small on tablet and large on desktop
+     * space = ['none', 'small','large']
```
afontcu, line 34:
```suggestion
     * :space="['none', 'small', 'large']"
```

src/components/MaStack/MaStack.vue
```diff
@@ -28,7 +42,15 @@ export default {
...
     /**
-     * Set the children alignment. Defaults to fill.
+     * Set the children alignment.
+     *
+     * If an array is passed, values will target the design system breakpoints.
+     * ```ts
+     * // This would apply left on all the different breakpoints
+     * align = 'left'
+     * // This would apply fill on mobile, center on tablet and left on desktop
+     * align = ['fill', 'center','left']
```
afontcu, line 52:
```suggestion
     * :align="['fill', 'center', 'left']"
```

src/components/MaTextField/MaTextField.stories.js
```diff
@@ -1,183 +1,114 @@
...
+const Template = (args, { argTypes }) => ({
+  data: () => ({ inputText: '' }),
+  props: Object.keys(argTypes),
+  template: `
+    <ma-text-field
+      v-model="inputText"
+      v-bind="$props"
+      :disabled="disabled"
+      :aria-label="ariaLabel"
+      :placeholder="placeholder"
```
afontcu, line 46: can we safely remove these lines since we're already binding `$props`?

vetur/tags.json
```diff
@@ -71,22 +71,22 @@
...
+    "description": "Renders a sidebar component following the Design System guidelines\n\n[Component's API documentation](https://holaluz.github.io/margarita/?path=/story/components-sidebar--sidebar)"
   },
   "ma-stack": {
     "attributes": ["space", "align"],
-    "description": ""
+    "description": "Renders a stack component following the Design System guidelines\n\n[Component's API documentation](https://holaluz.github.io/margarita/?path=/story/layout-stack--stack)"
   },
   "ma-text": {
     "attributes": ["tag", "size", "italic", "bold", "tone"],
     "description": "Renders text following the Design System guidelines"
```
afontcu, line 86: Looks Like MaText could use a DS link, too 😉

afontcu (commented):
neat 👌👌👌 left several comments here and there, but it looks great overall

src/components/MaHidden/MaHidden.js
edufarre: I will remove them. However, as the component is written into a `.js` file, vetur is no catching any of this, which means that we do not have IntelliSense when using MaHidden component`

src/components/MaDatagrid/MaDatagrid.vue
edufarre: I tried to do some work around here, setting its type to: `{import('vue').PropType<VNode>}`

src/components/MaTextField/MaTextField.stories.js
edufarre: I can remove two of them. When trying to remove `aria-label` in the story, it does set the attribute the following way:
Screen Shot 2021-02-17 at 10 12 20

Due to this issue, I removed `disabled` and `placeholder` but let `aria-label` there

sergigraciaj (approved):
Wow, good job!!!

src/components/MaTextField/MaTextField.stories.js
afontcu: lol, because aria-label is not a prop, it is an attribute 🤔 I guess we should actually define it as a prop, it makes way more sense XD

edufarre: Shall we tackle this into another PR? I do believe it might be a bit out of scope

afontcu: sure 👍

src/components/MaDatagrid/MaDatagrid.vue
afontcu: but it's not a PropType, right? I mean, it is just a key within an object, not a prop. Since we can't actually get type inference (or typechecking, FWIW), I wouldn't try to get its type if we cannot check its appropriate (and useful, since setting the wrong one won't make anything break, as far as I can tell)

edufarre: So.. shall I just delete this section? I do believe that having the example it might help when using the component. However, It can't be 100%, although I would just set the `component` prop as `any` then. The most important thing here, is that you can actually check which structure composes each of the columns' elements

afontcu: yeah, I guess we could keep it and set it to a _fake_ type that wouldn't compile, but might help. Something like `VueComponent` or whatever.

alexcastelo95 (approved):
Sorry for the delay, I was looking forward to see how this works calmly. I think it's super cool.

PR #361, opened 2021-02-22
fix: Remove MaSidebar

afontcu:
Question to @holaluz/frontend: how do you want to tag this release? It is clearly a breaking change, but it is already handled in https://github.com/holaluz/bonasera/pull/821, and won't affect any other projects. Issuing a release means manually bumping margarita to the next major version, so maybe avoiding the hassle is sensible. Thoughts?


edufarre:
I would go ahead and not release a new major version then, a minor would do, right?
However, I do reckon that we should bear in mind that in the PR where we update our storybook to use Controls, we do still have a Sidebar component, thus, I implemented its story https://github.com/holaluz/margarita/pull/360/commits/29951e3e6ccbfb29364deee055d60b5b296e425e, we must remove it from there too

sergigraciaj:
I'm ok with no releasing a major, but not sure how to tag it, maybe _perf_? do we have that? lol

afontcu:
> we should bear in mind that in the PR where we update our storybook to use Controls

I'd go ahead and remove the commit entirely from the PR no matter the release version that we end up issuing 👍

edufarre:
> > we should bear in mind that in the PR where we update our storybook to use Controls
> 
> I'd go ahead and remove the commit entirely from the PR no matter the release version that we end up issuing 👍

I know there is no relation with the discussed release version, I was just doing a reminder.
About the version, I would even say this is kinda `fix` isn't it? We are actually removing a non-used component. It don't even have a story nor documentation.

alexcastelo95 (approved):
I am still internalising the release tags. But since it's not being used anywhere, I think it's closer to a `fix` than anything else, right? 🤔

edufarre (approved):

PR #362, opened 2021-03-01
feat: Add layout components

### LINKED ISSUE #274: [RFC] Add Columns/Column component for vertical layout
Related to #256.

## Impact

- New `MaColumns` component.
- New `MaColumn` component.
- Deprecate `MaGridContainer`,  `MaGridRow`, `MaGridColumn`.

## Details

This RFC aims to include two-component set, with a simpler API, which can be used in conjunction with other layout components (Stack and other) and the responsive props proposed in #273.

## Proposed API

```vue
<ma-columns>
  <ma-column>content 1</ma-column>
  <ma-column>content 2</ma-column>
</ma-columns>
```

### Props for `MaColumns`

```js
space // layout-based space modifiers (small, medium, large…). Same as Stack.
collapse-below // responsive values (mobile, tablet, desktop)
align // align modifiers (start, center, left). Same as Stack. Useful in combination with collapse-bellow, and when columns do not reach 100% width.
align-vertical: "top" / "bottom" / "center". // Responsive prop (can receive an array of values)
```

`collapse-bellow` converts a column into a stack below the specified breakpoint. We could somehow replace MaColumn with MaStack (via `<component is="">`), but I believe this is such a common scenario it makes sense to provide a prop to solve this specific use case.

### Props for `MaColumn`

```js
width: "content" / "1/2" / "1/3" / "2/3" / "1/4" / "3/4"
```

If not specified, columns' width will be fluid, meaning they will take up to all available space:

```vue
<ma-columns>
  <ma-column width="1/3">...</ma-column>
  <ma-column>....</ma-column> <!-- it'll take two thirds of space -->
</ma-columns>
```

```vue
<ma-columns>
  <ma-column width="3/4">...</ma-column>

  <!-- both columns will split the remaining space (1/4) equally -->
  <ma-column>...</ma-column>
  <ma-column>....</ma-column>
</ma-columns>
```

`"content"` means a column that only takes as much space as needed:

```vue
<ma-columns>
  <ma-column width="content"><div style="height:30px;width:80px"></div></ma-column> <!-- 80px -->
  <ma-column>....</ma-column> <!-- 100% of width minus 80px -->
</ma-columns>
```

(Do we need fifths?)

## Examples

Collapsed below tablet with responsive space. Small gap on mobile (also collapsed), large otherwise.

Also, second column takes up two thirds of space.

```vue
<ma-columns collapse-below="tablet" :space="['small', 'large']">
  <ma-column>...</ma-column>
  <ma-column width="2/3">...</ma-column>
</ma-columns>
```

Columns width default (`medium`) spacing that become horizontally centered on mobile:

```vue
<ma-columns collapse-below="tablet" align="center">
  <ma-column>...<ma-column>
  <ma-column>...<ma-column>
  <ma-column>...<ma-column>
</ma-columns>
```

## Alternatives

Not developing this component at all, an rely on CSS Grid + layout space tokens to create our columns. We'd miss responsive props, but we'd win several degrees of freedom – not sure if a good or a bad thing XD

---

Thoughts? Ideas? Questions? Doubts?

afontcu:
This PR closes #274 

- [x] Port MaColumns
- [x] Port MaLayout
- [x] Add missing tests for Layout
- [x] Add vetur autocompletion prop tags
- [x] Add docs with examples
- [x] Document code where needed
- [ ] Also: right now, Columns and Layout use `gap` as a prop name, while Stack uses `space`. 
- [x] Check functionalDataMerge

^ any help with missing steps will be greatly appreciated! And woud help people understand the components better…

### DIFF
--- a/src/components/MaLayout/MaLayout.spec.js
+++ b/src/components/MaLayout/MaLayout.spec.js
@@ -1,0 +1,41 @@
+/* eslint-disable no-console */
+import { render } from '@margarita/margarita-test-utils'
+import MaLayout from './MaLayout'
+
+describe('MaLayout', () => {
+  test('adds only one stack if no columns are provided', () => {
+    const { stacks, grids } = renderComponent({ columns: '' })
+
+    expect(stacks).toHaveLength(1)
+    expect(grids).toHaveLength(0)
+  })
+
+  test('adds one stack and one grid if columns are provided', () => {
+    const { stacks, grids } = renderComponent({ columns: '1' })
+
+    expect(stacks).toHaveLength(1)
+    expect(grids).toHaveLength(1)
+  })
+
+  test('adds one stack and two grids if two rows are provided', () => {
+    const { stacks, grids } = renderComponent({ columns: '1 - 1' })
+    expect(stacks).toHaveLength(1)
+    expect(grids).toHaveLength(2)
+  })
+})
+
+function renderComponent(props) {
+  const utils = render(MaLayout, {
+    props: { gap: 'small', columns: '12', space: 'small', ...props },
+    slots: { default: 'content' },
+  })
+
+  const stacks = utils.baseElement.querySelectorAll('.stack')
+  const grids = utils.baseElement.querySelectorAll('.grid')
+
+  return {
+    ...utils,
+    stacks,
+    grids,
+  }
+}
--- a/src/components/MaLayout/MaLayout.spec.js
+++ b/src/components/MaLayout/MaLayout.spec.js
@@ -22,12 +22,32 @@ describe('MaLayout', () => {
     expect(stacks).toHaveLength(1)
     expect(grids).toHaveLength(2)
   })
+
+  test('keeps additional attributes', () => {
+    const { getByTestId } = render({
+      components: { MaLayout },
+      template: `
+        <ma-layout data-testid="layout" gap="small" columns="12" id="123" :class="['class1', 'class2']">
+          <div data-testid="div" id="456" class="class"></div>
+        </ma-layout>
+      `,
+    })
+
+    const layout = getByTestId('layout')
+    expect(layout).toHaveAttribute('id', '123')
+    expect(layout).toHaveClass('stack', 'class1', 'class2')
+
+    const div = getByTestId('div')
+    expect(div).toHaveAttribute('id', '456')
+    expect(div).toHaveClass('class')
+  })
 })
 
-function renderComponent(props) {
+function renderComponent(props, attrs = {}) {
   const utils = render(MaLayout, {
     props: { gap: 'small', columns: '12', space: 'small', ...props },
     slots: { default: 'content' },
+    ...attrs,
   })
 
   const stacks = utils.baseElement.querySelectorAll('.stack')
--- a/src/components/MaStack/MaStack.vue
+++ b/src/components/MaStack/MaStack.vue
@@ -1,4 +1,6 @@
 <script>
+/** @typedef {import('vue').VNodeData} VNodeData */
+import { mergeData } from 'vue-functional-data-merge'
 import { responsivePropValidator } from '@margarita/utils/responsivePropValidator'
 import { spacing } from '../../tokens'
 
@@ -66,22 +68,16 @@ export default {
     const space = spacing[responsiveSpace]
     const align = alignment[responsiveAlign]
 
-    return createElement(
-      'div',
-      {
-        class: {
-          stack: true,
-          ...data.class,
-        },
-        style: {
-          gap: space,
-          justifyItems: align,
-          ...data.style,
-        },
-        ...data,
+    /** @type {VNodeData} */
+    const componentData = {
+      staticClass: 'stack',
+      style: {
+        gap: space,
+        justifyItems: align,
       },
-      slots().default
-    )
+    }
+
+    return createElement('div', mergeData(data, componentData), slots().default)
   },
 }
 </script>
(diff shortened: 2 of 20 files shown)


Samitier:
Added documentation and examples for ma-layout in 554ce7a5b9567bd3fa87e00e4872c60ced0018d7 (sorry for the mess I did with the commits :trollface: ). Let me know if there's enough with that!

Also IMHO: we shouldn't document MaColumns in storybook since it should be of "internal" use only. Code will be more readable for everyone if we all create layouts with the same component instead of with multiple.

The gap-stack thing smells a bit, but lets not add a breaking change, no?

As for the missing tests,  if you are more specific I can try to add them too :smiley:

afontcu:
> Also IMHO: we shouldn't document MaColumns in storybook since it should be of "internal" use only. Code will be more readable for everyone if we all create layouts with the same component instead of with multiple.

I see your point. However, given the fact that we still expose Stack (and that it is really useful in some common scenarios) would it make sense to also expose its horizontal counterpart?

src/components/MaStack/MaStack.vue
```diff
@@ -22,6 +17,8 @@ const alignment = ['left', 'center', 'right', 'fill']
 export default {
   name: 'MaStack',
 
+  functional: true,
```
afontcu, line 20: Is this a potential breaking change? 🤔

Samitier:
> I see your point. However, given the fact that we still expose Stack (and that it is really useful in some common scenarios) would it make sense to also expose its horizontal counterpart?

The thing is... should we keep using stack? 😬 

```
<ma-stack space="small">
```
should be exactly the same as
```
<ma-layout gap="small">
```
so...why not use only `ma-layout`? Not saying that we should introduce a breaking change, just a "change mentions of stack to only use `ma-layout` from now on". Or is it a bit scummy?

src/components/MaStack/MaStack.vue
Samitier: hmm everything should work fine because all "parent" attributes seem to be covered in the render() function. Stuff like `data.staticClass` and `data.styles`. So things like:
`<ma-stack class="whatever" >` in theory should continue working as always. But yeah...a bit risky until we see it in action 😬

afontcu: I think it is currently breaking a MaModal test where Modal relies on a MaStack DOM node to add other attributes (such as `ref`). I'll take a look into it. We might need to instruct users not to use Layout components to set attributes other than their props to avoid weird scenarios.

Or take a look at [vue-functional-data-merge](https://github.com/alexsasharegan/vue-functional-data-merge), even though I'm not sure it'd fix all our issues

afontcu:
> `<ma-layout gap="small">`

To be quite honest I'm not entirely sure this API is as easy to understand as `<ma-stack space="small">` 😅

Samitier:
> To be quite honest I'm not entirely sure this API is as easy to understand as `<ma-stack space="small">` 😅

Aggree hahah.

We can always write the full thing: `<ma-layout gap="small" columns="12">`. Still not as clear as ma-stack  but at least we will "talk in 12 col layout". Also more flexible if , for example, we don't want a full-width stack `<ma-layout gap="small" columns="6">`.  

But I don't know, I don't think my opinion should count here because I'm full of bias in favor of ma-layout, so I'll stick on what you decide 😂

afontcu:
> We can always write the full thing: `<ma-layout gap="small" columns="12">`. Still not as clear as ma-stack but at least we will "talk in 12 col layout"

Yeah, that's true. Thoughts, folks? @holaluz/frontend

icurdu:
> Re Layout/Column/Stack debate, I'd suggest keeping things consistent:

For me, consistency is part of the core of the Design System. In order to achieve that, I think we should go all-in with MaLayout! I think giving one option will keep consistency up high, making everyone build the same way and talking with the same language. I still like MaStack tho, and we can keep it in our system, but I wouldn't expose MaColumn & MaStack.

src/components/MaStack/MaStack.vue
Samitier: aggree! IMHO this is super important if we want to keep using functional components too. We should be careful to "transfer" refs, static classes, dynamic classes, inline styles... I don't know the best way to do this. Maybe that dependency is what we need 🤔

Samitier:
I just added some examples for vertical align, because we forgot about it 😅 -> https://github.com/holaluz/margarita/pull/362/commits/cf5b56b98312f792a2833b55f44b2fb7afa2696d

src/components/MaStack/MaStack.vue
afontcu: I've added a couple of tests on both Layout and Slack to make sure they pass down some attributes. I've added vue-functional-data-merge just to make it more terse (and easier to reason about). Looks like we're safe!

Samitier (approved):

src/components/MaLayout/MaLayout.spec.js
```diff
@@ -0,0 +1,61 @@
...
+    })
+
+    const layout = getByTestId('layout')
+    expect(layout).toHaveAttribute('id', '123')
+    expect(layout).toHaveClass('stack', 'class1', 'class2')
+
+    const div = getByTestId('div')
+    expect(div).toHaveAttribute('id', '456')
+    expect(div).toHaveClass('class')
+  })
```
afontcu, line 43: 🎉 @Samitier

Samitier: nicee 🎉

PR #364, opened 2021-03-05
fix: Fix alignment when using MaText and input components

afontcu:
…to fix alignment when used in conjunction with other MaText components. Related to https://holaluz.slack.com/archives/CR9670R5H/p1614869238064300

- [x] Update MaTextField
- [x] Update MaSelect
- [x] Update MaRange

### DIFF
--- a/src/components/MaRange/MaRange.css
+++ b/src/components/MaRange/MaRange.css
@@ -5,7 +5,7 @@
   --range-bullet-size: 4px;
 
   position: relative;
-  margin-top: 4px;
+  margin-top: 8px;
   padding-bottom: 2rem;
 
   &::after {
@@ -16,10 +16,6 @@
     height: var(--range-height);
     content: '';
   }
-}
-
-.ma-range__label {
-  color: var(--color-gray-dark);
 }
 
 .ma-range__step {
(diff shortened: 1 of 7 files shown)


afontcu:
not sure if `feat` or `fix` 🤷‍♀️

edufarre (approved):

edufarre:
> not sure if `feat` or `fix` 🤷‍♀️

I do believe this is a `fix`, as it fixes (good redundancy here) the displayed text styles/appearance

afontcu:
> I do believe this is a `fix`, as it fixes (good redundancy here) the displayed text styles/appearance and we are not delivering any new feature or something that did not have before this PR, right?

One could say that the PR adds the ability to work in conjunction with MaText 😄 that being said, totally ok with a fix release 😉 updating the title

Moreover, let's talk about this in our next meeting. We could simplify things by only releasing minor versions and forgetting about the distinction between minor and patch

edufarre:
> One could say that the PR adds the ability to work in conjunction with MaText 😄 that being said, totally ok with a fix release 😉 updating the title

just gave my point of view, didn't say it was the correct one 😞  I did not think about the approach you said, although I reckon it makes 100% sense.

Issue #281, opened 2020-11-10
Make ma-button an anchor if paramerer href exists


Samitier:
We just had a bug in `dry-martini` where we had a `ma-button` that acted as an `anchor`.

We had something like this:

```html
<ma-button href="whatever">foo</ma-button>
```
And we should have had something like this:
```html
<ma-button tag="a" href="whatever">foo</ma-button>
```

This resulted on the button not doing anything.

I think this component's API allows this kind of errors and I'm sure this bug will happen again. 

Thats why I'm wondering why do we need the `tag` prop at all? Isn't it redundant? If the `ma-button` has a `href` prop, then it's an `anchor`. If not, it's a `button`. No need of adding redundant props, no?

What do you think?

afontcu:
Hum, how weird, I do recall having some MaButton with `href="/url" tag="a"` in bonasera that worked properly 🤔 

That being said, I agree that `tag` is redundant. I'd go the extra mile and say that, if we need "a button that looks like a link" or "a link that looks like a button", we could even develop those components and make sure they are accessible enough (it is quite common to mess things up when merging behavior (button, link) with UIstying (look like a button, look like text))

carlotamolintr:
Hi,
I agree with you that is redundant. Also at the beginning it happened to me to try to use Ma Button and was not working because I forgot to  write the tag...

> Hum, how weird, I do recall having some MaButton with `href="/url" tag="a"` in bonasera that worked properly 🤔
> 
> That being said, I agree that `tag` is redundant. I'd go the extra mile and say that, if we need "a button that looks like a link" or "a link that looks like a button", we could even develop those components and make sure they are accessible enough (it is quite common to mess things up when merging behavior (button, link) with UIstying (look like a button, look like text))

Regarding the above commented... I don't know what to say....

I will remove the tag but this will be a breaking change, right?

Thank you all

afontcu:
> It will remove the tag but this will be a breaking change, right?

Removing the `tag` prop is definitely a breaking change, but adding the "right" components to handle "buttons that look like links" and "links that look like buttons" would not – we could add them, and then discuss about the breaking change. We could:

1) Remove `tag` – breaking change.
2) Add a warning (on dev env) warning that using `tag` on a Button is deprecated and likely to be removed in a future release. This would help us detect usages (and fix them) without the struggle of a breaking change.

Thanks for your thoughts!

edufarre:
> We could:

[...]

Issue #257, opened 2020-05-08
Add layout components for the new Design System.


Samitier:
## What

Add a horizontal/grid layout component that allows us to stack and space UI elements using the spacing properties of the Design System:

![image](https://user-images.githubusercontent.com/4160121/81422720-a680e680-9153-11ea-9390-301ef8afbcb7.png)

Also that it respect container requisites of the design system:

![image](https://user-images.githubusercontent.com/4160121/81423236-7ede4e00-9154-11ea-9e6c-da4c18c305bc.png)

## How (let's discuss it)

I would just do this with an external library to not reinvent the wheel to be honest. Lets just use some layout system that is already proven to work and tweak it so we can use it with the params of our Design System. I don't think its necessary to use `grid` if we combine it with the stack component we will already have.

## Tasks

- [ ] Decide the "how".
- [ ] Implement new approach.
- [ ] Document it.
- [ ] Implement figma layout pages in Storybook.
- [ ] Fix breaking changes (see below)

## Breaking changes

No breaking changes if we do not touch `<ma-grid>, <ma-colum>, etc`. Extremely complicated breaking changes if we do.

Samitier:
Closed by https://github.com/holaluz/margarita/issues/274

PR #367, opened 2021-03-09
feat: Add MaList component

sergigraciaj:
https://www.notion.so/holaluz/a1cb451d360941d98f08831cf679edab?v=f7024af195d34caface8649e58f98c12&p=238dd2b8cdf54575ac7bfaefd315e148

image


src/components/MaTextList/MaTextList.vue
```diff
@@ -0,0 +1,101 @@
+<template>
+  <component :is="tag" :class="iconClass">
+    <li
+      v-for="(listItem, index) in listItems"
+      :key="index"
+      :style="computedStyle"
+    >
+      <ma-text tag="p" :size="size" :tone="tone">
```
edufarre, line 8: Not sure about hardcoding `p` and then allowing through the API set the `size` and `tone` of the `MaText`, shall we allow to change the `MaText tag` ?

docs/components/MaTextList.docs.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+## How to use it? (Controls)
+
+- **Tag:** Set the HTML element tag to ensure the document semantics are meaningful
+    - ul → To present content of equal status or value.
+    - ol → Imply sequence and order, and are commonly used when giving a set of instructions.
+- **Size:** Sets the text element size according to our Design System
+    - MaText-Medium, MaText-Small
+    - The use of any MaHeading or MaText-xSmall is strictly forbidden.
+- **Tone:** Sets the text element color tone (check text-tones). This also applies to symbols of the bullet lists
```
edufarre, line 13: Is not really related to this PR, but, shall we write down a `tones` doc in order to set a link of the different props that depend on it, this sentence included?

edufarre (commented):
good job! 🚀

src/components/MaTextList/MaTextList.vue
afontcu: This feels a bit like a rabbit hole – that is, we might need to change the MaText wrapper component juuust a tiny bit, and then that other tiny bit, and then… I imagine a situation where we need to pass a whole component, because it has a special behavior: for instance, a link or a special color.

What if, and this is just a suggestion, we left the decision to the consumer, instead of trying to find the right abstraction for every use case?

```vue
<ma-list>
  <ma-text tone="red">Whatever</Text>
  <ma-text>Whatever 2</Text>
  <my-very-special-text-component text="Whatever 3" />
</ma-list>
```

With a prop we wouldn't be able to do this, either:

```vue
  <ma-list space="large">
    <ma-stack space="medium">
      <ma-text tone="gray-darker">A title</ma-text>
      <ma-text>A content</ma-text>
    </ma-stack>
    <ma-stack space="medium">
      <ma-text tone="gray-darker">A title</ma-text>
      <ma-text>A content</ma-text>
    </ma-stack>
  </ma-list>
```

sergigraciaj: Right now we only have use cases for lists with p, so if in a future we need a list of words we can modify that!

edufarre: But that would be a Breaking Change, right? I do not see the pain in iterating this now and making it a bit more flexible. What about if some of the list elements have a link to an external page or something like that?

src/components/MaTextList/MaTextList.spec.js
```diff
@@ -0,0 +1,64 @@
...
+  test('renders item from list', () => {
+    const { getByText } = renderComponent()
+
+    expect(getByText('Item 1')).toBeInTheDocument()
+  })
+
+  test(`'tag' property renders the valid provided tag element`, () => {
+    const { getByText } = renderComponent({ tag: 'ol' })
+
+    expect(getByText('Item 1').parentNode.parentNode.nodeName).toBe('OL')
```
afontcu, line 14: Both `ul` and `ol` have a role of `list`,  so you could use `getByRole('list')` instead

cillascillas:
Regarding comments [here](https://github.com/holaluz/margarita/pull/367#discussion_r591184888) and [here](https://github.com/holaluz/margarita/pull/367#discussion_r591100551) and after discussing them in DS meeting, there are some updates:

- List elements are HTML elements passed as slot. Each node is transformed in a list item. (So now we can add links, or more complex text elements)
- As list elements are expected/recommended to be `MaText` properties tone and size (from MaTextList component) are passed to `MaText` as well. However can be overwritten if defined in the component to enable customization.
- This behaviour info has been added to the docs tab info and 3 different examples of list items are now available in the controls tab.
- An error is logged if no HTML elements/list items are defined

src/components/MaTextList/MaTextList.spec.js
cillascillas: So true! Updated!

src/components/MaTextList/MaTextList.css
```diff
@@ -0,0 +1,57 @@
...
+  list-style: none;
+}
+
+li {
+  position: relative;
+}
+
+li p {
+  max-width: 75ch;
+}
```
afontcu, line 12: I'm not entirely sure a UI component should set a width limit – this is on Layout components to define

src/components/MaTextList/MaTextList.vue
```diff
@@ -0,0 +1,105 @@
...
+      functional: true,
+      render(createElement, { slots, props, data }) {
+        if (!slots().default) {
+          // eslint-disable-next-line no-console
+          console.error(`[TextList component] No list items found`)
+          return
+        }
+
+        return slots().default.map((item) => {
+          if (item.componentOptions && item.componentOptions.propsData) {
```
afontcu, line 27: Are we 100% sure we need to take `propsData` into account? If I'm not mistaken, [`propsData`](https://vuejs.org/v2/api/#propsData) is (was?) helpful when it came to unit testing, as it helped test utils to set initial prop values.



that being said, if this is still needed (could be), you could benefit from using the optional chaining and remove the if statement:

```js
item.componentOptions.propsData = {
  ...props,
  ...item.componentOptions?.propsData,
}
```

src/components/MaTextList/MaTextList.vue
```diff
@@ -0,0 +1,105 @@
...
+
+<script>
+import { text, tones } from '../../tokens'
+
+/**
+ * Renders text-list following the Design System guidelines
+ *
+ * [Component's API documentation](https://holaluz.github.io/margarita/?path=/story/components-textlist--text-list)
+ */
+export default {
```
afontcu, line 15: Looks like we're missing `name`.

With that… what if we named it `MaList`? Since we're using slots we're not tied to text, and its name may be confusing.

src/components/MaTextList/MaTextList.css
sergigraciaj: The width is defined by the parent, but the text is defined to have a max-width of 75ch
https://www.notion.so/holaluz/Text-Lists-238dd2b8cdf54575ac7bfaefd315e148

afontcu: > The width of the <li> element is defined by its parent but has a max-width of 75ch to have proper legibility.

I assume this is true for any text-based component, and we're not setting a max-width on `MaText` right (because it is not its goal)?

src/components/MaTextList/MaTextList.vue
sergigraciaj: I'm not sure about that, MaTextList shows the component intention, a list of texts. MaList can be confusing, it may seems like it can contain not just texts (as Stepper will do).

afontcu: sure! your call, resolving this 👍 

Just mentioned it because we'd also have a lot of "Text" components: MaText, MaTextField, and now MaTextList. But I guess we'll get used to it

src/components/MaTextList/MaTextList.css
sergigraciaj: Yup, it should be applied to any `li` direct child, so if the text slot is made of different pieces (span + a or double span) it should be contained inside a parent div.
And then this would do the job
```
li > * {
  max-width: 75ch;
}
```

Samitier:
Sorry for being a bit late to the game:

Aren't the props `tag` and `icon` conflicting? I mean, what happens if it's an `ol` with an icon? We have 3 states: ordered list, bullet list and check list. Why do we need 2 props for that? Feels like https://github.com/holaluz/margarita/issues/281 again.

IMHO, we should have a prop `type` with 3 states (`ordered`, `bullet`, `check`) and that's it. 

If you already discussed this, forget what I've said.

Also, a bit offtopic: what happens with the other `ol` we have (dry/maitai)?

image

src/components/MaTextList/MaTextList.css
afontcu: I might have not explained myself properly 🙏  What I meant is that we have other text-based components (MaText) and we're not setting a max-width, mostly because it is out of the scope for a UI component (margin and sizing should be done using Layout components, not UI ones). So this rule here introduces friction between both sets of components.

You'd be using a UI component, and out of the blue it'd set a max width for its internal content. That's a surprising effect to me

src/components/MaTextList/MaTextList.vue
cillascillas: Asked to @icurdu and he turned it[ into a poll in DS slack channel](https://holaluz.slack.com/archives/C019DLZTBU6/p1615799625024300), let's wait for the result!

src/components/MaTextList/MaTextList.vue
cillascillas: In this case it is used to pass props to the element we are creating. However, if there's any other way to do it, I'm all 👂 👂 !

afontcu: Not really sure, but if it works, it works 😉

src/components/MaTextList/MaTextList.css
cillascillas: Not sure if I'm getting what you're discussing, but if I'm not wrong we set the max width for the component according to [this design definition of the component](https://www.notion.so/holaluz/Text-Lists-238dd2b8cdf54575ac7bfaefd315e148#1021c7af7c0a46769a75da6f94e86460), does it make sense?

afontcu: yup saw that, what I mean is that UI components (such as MaList or MaText or MaCard or MaButton) should not be responsible for setting widths/margin/paddings, because this is why Layout components exist. At least, this was [the initial vision](https://github.com/holaluz/margarita/issues/257#issuecomment-635925152) when we introduced Layout components.

Letting MaList set a max width is stepping outside of its boundaries: you might place a List inside a Stack or a Layout, and it will behave differently than other UI components.

src/components/MaTextList/MaTextList.css
```diff
@@ -0,0 +1,57 @@
...
+  margin-bottom: 0.5rem;
+}
+
+ul li::before {
+  content: '';
+  position: absolute;
+}
+
+ul.bullet li::before {
+  background: url('../../assets/icons/bullet.svg');
```
Samitier: why do we need an svg to render a bullet point?

Is `content: '•'` not enough?

src/components/MaTextList/MaTextList.css
```diff
@@ -0,0 +1,57 @@
...
+  counter-reset: counter;
+}
+
+ol li::before {
+  counter-increment: counter;
+  content: counter(counter) '.';
+  position: absolute;
+  top: 0;
+  left: -1.25rem;
+}
```
Samitier: isn't all this the same as:

```suggestion
ol {
  list-style-type: decimal;
}
```

src/components/MaTextList/MaTextList.css
carlotamolintr: Relate with that, now the bullet size is not changing depending on the size prop selected. Do we want to change the bullet size? I don't know...

carlotamolintr:
It looks good!!👏👏👏

src/components/MaTextList/MaTextList.css
cillascillas: Let me see if I got this straight, I guess that what you mean is that we shouldn't set this max width here, and that it should be something that the design team takes into account when designing. And then, it will be "shaped" by Layout or Stack when implemented. Am I getting it right?

afontcu: yeah, exactly 😃 – a List component should render a List, whereas a Layout component should help laying out other components. Having a List component set a max-width should look as weird as having a Layout component defining the default `ul` element styles (to give an example).

cillascillas: Ok, just removed it and wrote a reminder for the design team to take it into account!

cillascillas:
According[ to this comment](https://github.com/holaluz/margarita/pull/367#issuecomment-799214421), I really think that this is a great approach:

> IMHO, we should have a prop type with 3 states (ordered, bullet, check) and that's it.

So, it's been revisited and working properly 🤞 . Docs have been updated as well and I'm asking the design team to review it!

src/components/MaTextList/MaTextList.css
cillascillas: If I recall correctly we chose the `::before` in order to custom the spacing (because we couldn't manage to do it with `::marker`). However, after mentioning it here, maybe it's better for instance in terms of vertical alignment. So, let's use `list-style-type: decimal;` and discuss with design about it.

src/components/MaTextList/MaTextList.css
cillascillas: Initially we chose the _svg_ approach just to be consistent and have both of the icons the same way (Also because we wanted to avoid playing with the `font-size`), but it's something we can reconsider and revisit in the future when we see if there are more types to be added of if we use them.
Regarding the size, I checked the [figma file](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/Design-System-Beta?node-id=0%3A1) and the size of the bullet/icon is the same for both options (small & medium).

docs/components/MaTextList.docs.mdx
cillascillas: Sounds good! I created [this card](https://holaluz.kanbanize.com/ctrl_board/60/cards/18663/details/) so we don't forget about it!

src/components/MaTextList/MaTextList.vue
cillascillas: Discussed in DS meetings and implemented! Thanks for the suggestion!

cillascillas:
#### Updates after review:

- CSS improvements (including unset max-width for list item)
- Renaming MaTextList to MaList
- API: Replacing `tag`/`icon` props to new type prop with options: `bullet`, `check`, `ordered` and update docs to contemplate it.

Samitier (approved):

src/components/MaList/MaList.vue
```diff
@@ -0,0 +1,103 @@
...
+          return
+        }
+
+        return slots().default.map((item) => {
+          if (item.componentOptions && item.componentOptions.propsData) {
+            item.componentOptions.propsData = {
+              ...props,
+              ...item.componentOptions?.propsData,
+            }
+          }
```
afontcu, line 34: If the if statement needs to stay, then the optional chaining operator is redundant 😇

afontcu (approved):
LGTM!

src/components/MaList/MaList.vue
cillascillas: Yes! Sorry, I was testing it and I forgot to remove it! Thanks for the 👀 !

Issue #366, opened 2021-03-09
MaModal is not responsive and outlines links


Samitier:
MaModal is not responsive for mobile (or screens with a small height):

![image](https://user-images.githubusercontent.com/4160121/110504866-e14e0500-80fd-11eb-9838-6e73e10a84ce.png)

And the focus trap we implemented focuses on links and gives them an outline when you open the modal, which imho is pretty weird.

afontcu:
> And the focus trap we implemented focuses on links and gives them an outline when you open the modal, which imho is pretty weird.

That's the expected behavior for a Modal [given the spec](https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-7), to focus on the first focusable element of its content 🤔 

> When a dialog opens, focus moves to an element inside the dialog. See notes below regarding initial focus placement.

 
![image](https://user-images.githubusercontent.com/9197791/110598488-18fe9080-8182-11eb-874a-5d23d74cdfb7.png)

Samitier:
Yeah, we will make the outline CSS less intrusive in maitai/dry and that's it.

afontcu:
> Yeah, we will make the anchor's outline CSS less intrusive in maitai/dry and that's it.

That sounds reasonable 👍 we could also think about letting consumer to define which element has the initial focus (to make it able to fulfill points 3, 4, and 5 of screenshot above).

Issue #357, opened 2021-02-11
MaModal allows body to overflow and scroll


afontcu:
We're not controlling body overflow when Modal is opened, so user can scroll its content. I believe this should not happen.

PR #368, opened 2021-03-10
fix: MaModal responsiveness, UX & styling issues

Samitier:
- Fixes https://github.com/holaluz/margarita/issues/366
- Fixes https://github.com/holaluz/margarita/issues/357
- Uses MaHeading for the Heading.
- Fixes focus trap bug when there are no focusable elements inside ModalContent.
- The modal header white has now less padding because it was a bit ugly.\
- Modifies the close icon a bit so it looks more like the one in figma

### DIFF
--- a/src/components/MaModal/MaModal.vue
+++ b/src/components/MaModal/MaModal.vue
@@ -28,7 +28,13 @@
               class="modal-header"
               data-testid="modal-header"
             >
-              <span class="modal-title">{{ title }}</span>
+              <ma-heading
+                class="modal-title"
+                :tone="headerType === 'gradient' ? 'white' : 'gray-dark'"
+                size="xsmall"
+              >
+                {{ title }}
+              </ma-heading>
               <ma-button
                 category="no-background"
                 data-testid="close-button"
@@ -56,6 +62,7 @@
 import { Portal as MaModalPortal } from '@linusborg/vue-simple-portal/dist/index.umd'
 import MaIcon from '@margarita/components/MaIcon'
 import MaButton from '@margarita/components/MaButton'
+import MaHeading from '@margarita/components/MaHeading'
 
 const FOCUSABLE_ELEMENTS = [
   'button',
@@ -81,6 +88,7 @@ export default {
     MaModalPortal,
     MaIcon,
     MaButton,
+    MaHeading,
   },
 
   props: {
--- a/src/components/MaModal/MaModal.vue
+++ b/src/components/MaModal/MaModal.vue
@@ -9,17 +9,16 @@
     </div>
     <ma-modal-portal>
       <transition name="modal-transition" @after-leave="closeModal">
-        <div v-if="showModal" class="modal-wrapper">
+        <div
+          v-if="showModal"
+          class="modal-wrapper"
+          data-testid="overlay"
+          @click.self="closeModal"
+        >
           <div
-            class="modal-overlay"
-            data-testid="overlay"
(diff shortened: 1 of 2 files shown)


Samitier:
Before I start a war, I removed the `ma-stack` because I was unable to make the modal-content `overflow:auto` with a `display: grid`. I don't know if it is a limitation of `grid` or a limitation of my brain but I only could do it with `display: flex` and `flex: 1`.

afontcu:
Looks a bit off to me when used on a really short screen: 

image


Not sure if we should support those, thou…


> I was unable to make the modal-content overflow:auto with a display: grid

No idea 😂 so if this really fixes the issue and nobody is able to come up with a better solution with Stack, I'm OK with it xD

Samitier:
> Looks a bit off to me when used on a really short screen:

Lets talk about that massive margin-top in the design system today hahah maybe we should add a media query. I think that as long as we support iphone 5, we should be ok.

src/components/MaModal/MaModal.vue
```diff
@@ -173,6 +180,7 @@ export default {
        */
       this.$emit('close')
 
+      document.body.style.overflowY = ''
```
Samitier, line 183: should we store the "before" state of overflowY and set it here? In case there was already an inline overflowY in the body and we remove it here. Or is it so edge case that we can ignore it?

afontcu: damn this is nasty but I believe it is the only way it'll work other than creating a global class for `body` xD

I'd store and reset the default value just to avoid future footguns… it's an edge case but we're already here so why not avoiding it :)

src/components/MaModal/MaModal.vue
```diff
@@ -191,6 +199,8 @@ export default {
 
       if (firstFocusableElement) {
         firstFocusableElement.focus()
+      } else if (this.$refs.closeButton) {
+        this.$refs.closeButton.$el.focus()
```
afontcu, line 203: I'm still not sure why this is working properly 🤔  I mean, `ref="modal"` is crearly set to the Modal wrapper, which includes the close button. `this.focusableElements` should always, at least, contain the closing button… weird.

src/components/MaModal/MaModal.vue
Samitier: well, we could create a class and set it here and it wouldnt be necessary to store the previous state of style.overflowY

that class can live inside an unscoped css in this component. Whatever you guys prefer, both options are shady anyways 😂

src/components/MaModal/MaModal.vue
Samitier: I think the problem starts when there are focusable elements outside the modal. If there are no focusable elements outside, the focus trap "works" because we return the focus inside the modal. I don't know if that is what makes the test work regardless

afontcu: > I don't know if that is what makes the test work regardless

Oh, nice catch. Let me give it a go and come back with an answer 👍

src/components/MaModal/MaModal.vue
afontcu: sure, maybe `document.body.classList.add('no-overflow')` (and creating a global `.no-overflow` class on a global CSS file) looks "cleaner"… but yeah, still nasty 😂

afontcu (approved):

carlotamolintr (approved):

PR #371, opened 2021-03-11
feat: Add suffix to MaTextfield

sergigraciaj:
https://www.notion.so/holaluz/TextField-with-suffix-07e64236dfbe4ef190c72756074ca46d


src/components/MaTextField/MaTextField.vue
```diff
@@ -21,6 +21,15 @@
         class="ma-text-field__input"
         v-on="inputListeners"
         @keyup.enter="removeFocus"
+        @focus="addFocus"
```
afontcu, line 24:
```suggestion
```

sergigraciaj: Oopsie, thx!

src/components/MaTextField/MaTextField.css
```diff
@@ -17,36 +17,52 @@
 
 .ma-text-field__input-wrapper {
   display: flex;
+  align-items: center;
   height: 2.25rem;
+  position: relative;
```
afontcu, line 22: I'm not 100% sure this is needed in this case:

```suggestion
```

afontcu (commented):
LGTM! just a couple of comments:

1. Notice that the input grows larger when we set a `suffix`. While input's length is something that should not depend on the input but on the parent (see MaLayout), just bear in mind that as of now, inputs with suffix will be larger than inputs without it. This might yield broken UIs, so we might want to find a workaround.

2. It looks a bit weird when input has an error:
image

also when errored + disabled: 

image

sergigraciaj:
Input with error solved, thx 👍 

> 1. Notice that the input grows larger when we set a suffix. While input's length is something that should not depend on the input but on the parent (see MaLayout) and should be eventually fixed, just bear in mind that as of now, inputs with suffix will be larger than inputs without it. This might yield broken UIs, so we might want to find a workaround.

This will be solved with MaLayout, right?
I also wonder if there are accessibility implications. 
We can check it in a future iteration.

afontcu:
> This will be solved with MaLayout, right?

Hm, have you tried it? It keeps growing larger after a fix width it might overflow Layout components. 

>  I also wonder if there are accessibility implications.

Not sure I follow. You mean about the input growing larger? I don't think there will be any.

Samitier:
Just one thing: shouldn't the suffix focus the input when clicked? 

It was literally the first thing I tried: clicked the suffix and started writing and nothing happened 😂

sergigraciaj:
> Hm, have you tried it? It keeps growing larger after a fix width it might overflow Layout components.

I tried it with a fix width parent and it doesn't overflow. Anyways we can revisit the component in a future if needed.

> Not sure I follow. You mean about the input growing larger? I don't think there will be any.

No, I mean the reader reading the suffix. I'm not sure if right know the reader will read the suffix after the user enters the value in the input, so I don't know if the suffix has to live before the input in the html (and then do a flex-direction: row-reverse). Or if using the suffix as a label will do the job.

> Just one thing: shouldn't the suffix focus the input when clicked?

Using the suffix as a label will do the job. Using two labels for an input is a problem?

Samitier:
> Using the suffix as a label will do the job. Using two labels for an input is a problem?

I mean this: 
image

The input hovers, but clicking on it does nothing.

sergigraciaj:
> The input hovers, but clicking on it does nothing.

That's why I say that using the suffix as label will do the job, if suffix is a label, then clicking it focus the input, but we will have two labels.

afontcu:
> That's why I say that using the suffix as label will do the job, if suffix is a label, then clicking it focus the input, but we will have two labels.

I think you could [per spec](https://html.spec.whatwg.org/multipage/forms.html#the-label-element), but:


"If you unintentionally associate multiple labels with a single control, you'll create inconsistent behavior across browsers and assistive technologies: some will read the first label, some will read the last label, and others will read both labels"

https://web.dev/form-field-multiple-labels/
https://scottaohara.github.io/tests/html-inputs/double-label.html

sergigraciaj:
Can we merge this PR as we want to use this component and discuss this in the closed PR and do a refactor then guys?

Samitier:
hahah I'm sorry but I still don't understand what the label has to do with anything 😂 

I meant just doing something like this:

```vue
 <ma-text
   v-if="suffix"
   @click="$refs.input.focus()" <- this
   size="small"
   tag="span"
   tone="gray"
   class="ma-text-field__input-suffix"
   v-text="suffix"
/>
```
but whatever, we can do this later if someone complains!

Samitier (approved):

afontcu:
> Can we merge this PR as we want to use this component and discuss this in the closed PR and do a refactor then guys?

sure 👍  it's not a refactor, though

afontcu (approved):

sergigraciaj:
`@click="$refs.input.focus()"` Is not working for me, we can check it in a future!

PR #373, opened 2021-03-16
fix: Fix MaOption vertical alignment

### LINKED ISSUE #346: MaOption requires a "small" prop
Kalimotxo uses it, and there's no way to vertically center it right now (because we're [hard setting a line-height](https://github.com/holaluz/margarita/blob/master/src/components/MaOption/MaOptionCheckbox.css#L20))

See:

https://github.com/holaluz/kalimotxo/pull/47/files#diff-5d43d5b3237eaa2a49886786010b464c1d0c5a145950eb62a3dc60327c820409R45-R49
https://github.com/holaluz/kalimotxo/pull/47/files#diff-c2a92a8b80760bc680301d0de8cc88f7c7f3c8f0dcfcd831f6807a9621749599R57

image

edufarre:
Closes #346 

After gathering a bit of info, I do believe that having a property that allows the `MaOption` to be 'small' does not have lot of sense, because looking at the Design System, we only have **one approach** of the Checkbox and Radio input elements.

[Radio styles in the DS](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/Design-System-Beta?node-id=3540%3A6520)
[Checkbox styles in the DS](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/Design-System-Beta?node-id=3540%3A6348)

This being said, it kinda makes sense to just one approach, the one that we have in our DS (which basically means that the text we use as description should be `small` sized)

Checkbox:
Screen Shot 2021-03-16 at 08 11 04

Radio:
Screen Shot 2021-03-16 at 08 11 11

Example with a multi-line checkbox:
Screen Shot 2021-03-16 at 08 18 58

### DIFF
--- a/src/components/MaOption/MaOption.vue
+++ b/src/components/MaOption/MaOption.vue
@@ -10,9 +10,9 @@
       v-bind="$attrs"
     />
     <span class="indicator" />
-    <span class="description">
+    <ma-text size="small" class="description">
       <slot />
-    </span>
+    </ma-text>
   </label>
 </template>
 
--- a/src/components/MaOption/MaOptionCheckbox.css
+++ b/src/components/MaOption/MaOptionCheckbox.css
@@ -4,7 +4,7 @@
 
 .ma-option--checkbox .indicator {
   position: absolute;
-  top: 2px;
+  top: 0;
   left: 0;
   transition: border 0.4s, background-color 0.33s;
   border: 1px solid var(--color-gray-light);
@@ -16,8 +16,7 @@
 .ma-option--checkbox .description {
   /* .indicator is absolute-positioned. We need to make room for the text */
   padding-left: calc(var(--indicator-size) + 0.5rem);
-
-  line-height: 1.3;
+  margin-top: 2px;
   color: var(--color-gray-dark);
 }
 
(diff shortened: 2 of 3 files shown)


vetur/attributes.json
```diff
@@ -72,7 +72,7 @@
   "ma-columns/justify": {
     "type": "string",
     "description": "Defines how the space between and around content items is distributed. Defaults to start.",
-    "options": ["space-around", "space-between", "start", "end"]
+    "options": ["space-around", "space-between", "start", "end", "center"]
```
afontcu, line 75: Not sure how this slipped through the cracks (my bad), but since MaColumns is an internal component we could simply remove all docs for it💡

afontcu (approved):
works for me! Question: have you checked that this looks good on kalimotxo's `CallMe`? I remember spending some time getting it right there (it was the use case that triggered the initial issue).

alexcastelo95 (approved):
We watched it together yesterday so I have all the context.
LGTM. 👍🏽

vetur/attributes.json
edufarre: I thought it was a bit weird, but as I wasn't certain I just let it there. I'll remove it from the docs, you're right

edufarre:
> works for me! Question: have you checked that this looks good on kalimotxo's `CallMe`? I remember spending some time getting it right there (it was the use case that triggered the initial issue).

This is what I'm about to do. Fix the `CallMe` in Kalimotxo.

PR #374, opened 2021-03-16
fix: Fix MaList vetur intellisense

sergigraciaj:


afontcu (approved):

edufarre:
I do think `fix: Fix MaList vetur intellisense` would be a more accurate commit message because you are not 'fixing vetur'

edufarre (approved):

sergigraciaj:
> I do think fix: Fix MaList vetur intellisense would be a more accurate commit message because you are not 'fixing vetur'

True, thanks!

PR #376, opened 2021-03-16
fix: scoped styles in layout

Samitier:
This fixes a bug where, by setting a class with scoped styles to a layout component, the css doesn't work.

This happened because if we have a scoped css in the parent component and also in the functional component, the data attribute of the HTML element gets overwritten by the css in the functional component.

Because of this, all parent scoped css stops working.

The solution? Stop using scoped css in functional components.

### DIFF
--- a/src/components/MaColumns/MaColumns.vue
+++ b/src/components/MaColumns/MaColumns.vue
@@ -91,7 +91,7 @@ export default {
 
     /** @type {VNodeData} */
     const componentData = {
-      staticClass: `grid`,
+      staticClass: 'ma-grid',
       class: {
         'has-auto-flow': hasAutoFlow({ columns }),
         [`vertical-align-${props.verticalAlign}`]: true,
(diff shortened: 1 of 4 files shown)


afontcu (approved):
Nice catch!

I'm _**really**_ worried that this might help/suggest people that overriding Layout component styles is encouraged 😅 so, just in case:

**It. Is. Not.** Please don't override styles for Layout components (or any other Margarita component, FWIW).

edufarre (approved):

edufarre:
> Nice catch!
> 
> I'm _**really**_ worried that this might help/suggest people that overriding Layout component styles is encouraged 😅 so, just in case:
> 
> **It. Is. Not.** Please don't override styles for Layout components (or any other Margarita component, FWIW).

I couldn't agree more. Actually, I just realized about this issue, because all the `ma-stack` components I did have with padding, stopped looking as expected.

afontcu:
Could we set an `!important` to all styles? I'm not sure if I'm kidding or not

Samitier:
> Could we set an `!important` to all styles? I'm not sure if I'm kidding or not

We could set everything as a :style too, so that no class can overwrite it (like it's done in maText)

Samitier:
> Could we set an `!important` to all styles? I'm not sure if I'm kidding or not

Honestly, I vote for it hahah. Imagine for a second the madness of debugging a layout component with overwritten css 😂

PR #378, opened 2021-03-19
feat: Add horitzontal align prop to ma text

carlotamolintr:


vetur/attributes.json
```diff
@@ -413,6 +413,11 @@
       "gray"
     ]
   },
+  "ma-text/align": {
+    "type": "string",
+    "description": "Set an horitzontal alignment to the text",
```
afontcu, line 418:
```suggestion
    "description": "Set an horitzontal alignment to the text. Only applies to block-level elements",
```

src/components/MaText/MaText.stories.js
```diff
@@ -32,9 +38,10 @@ export default {
 
 const Template = (args, { argTypes }) => ({
   props: Object.keys(argTypes),
-  template: `<div style="display: flex;">
+  template: `
+  <div style="width: 20rem">
```
afontcu, line 42: I'd keep the flex layout for now, so we can easily see the text without the blueish background

```suggestion
  <div style="display: flex;width: 20rem">
```

afontcu (approved):
LGTM! 🚀  I've added a couple of small suggestions

afontcu:
Remember that this PR is an addition, so its title (and thus the merge commit) should start with `feat: ` to create a new release

edufarre (commented):
Shall we add a test that ensures that given an `align` prop value, the expected alignment style is set?

carlotamolintr:
> Shall we add a test that ensures that given an `align` prop value, the expected alignment style is set?

Sure! I just did it. 🙂

edufarre (approved):

PR #381, opened 2021-03-24
docs: Adds docs for text field suffix

sergigraciaj:


docs/components/MaTextFieldWithSuffix.docs.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+👉 The user needs to enter an amount and whatever the amount → it will be € → so we fill the unit for them.
+
+## Variants
+
+[https://www.figma.com/embed?embed_host=notion&url=https%3A%2F%2Fwww.figma.com%2Ffile%2F7RhweKibaUnxX6WYBl7zKW%2FDesign-System-Beta%3Fnode-id%3D2572%253A1707](https://www.figma.com/embed?embed_host=notion&url=https%3A%2F%2Fwww.figma.com%2Ffile%2F7RhweKibaUnxX6WYBl7zKW%2FDesign-System-Beta%3Fnode-id%3D2572%253A1707)
+
+## Behavior
+
+Same as text-field.
```
edufarre, line 43: Could we add the link to our `TextField` documentation?

docs/components/MaTextFieldWithSuffix.docs.mdx
```diff
@@ -0,0 +1,51 @@
+# TextField-with-suffix
```
edufarre, line 1:
```suggestion
# TextField with a suffix
```

docs/components/MaTextFieldWithSuffix.docs.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+## Description
+
+Input field with independent (from the value) and fixed (always visible) suffix. 
+The suffix itself can be either text, symbols or both.
```
edufarre, line 6:
```suggestion
The suffix property accepts both text and symbols.
```

docs/components/MaTextFieldWithSuffix.docs.mdx
```diff
@@ -0,0 +1,51 @@
+# TextField-with-suffix
+
+## Description
+
+Input field with independent (from the value) and fixed (always visible) suffix.
```
edufarre, line 5: Not sure this sentence makes sense 🤔  I would change it to something like:
`Input field with a suffix section which is always displayed/visible `

docs/components/MaTextFieldWithSuffix.docs.mdx
```diff
@@ -0,0 +1,51 @@
...
+We can edit the same items from the text-field + the **suffix**.
+
+- Aria label
+- Edit label
+- Edit placeholder value
+- Add and edit suffix
+- Error message
+- State
+
+If the user misses that the input already has a suffix and enters it → Allow this in our validation and do not show an error.
```
edufarre, line 19: Shall we add this validation into our component in an incoming iteration?

docs/components/MaTextFieldWithSuffix.docs.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+⚠️  **How not to use it**
+
+- Do not rely only on suffixes, some users will not read them.
+- Suffixes will not replace labels. If you need a specific type of information, say so in the input label or hint/helper text as well.
```
edufarre, line 24:
```suggestion
- Suffix property does not replace the label's responsibilities. If you need a specific type of data, empathise it in the input's label or placeholder.
```

docs/components/MaTextFieldWithSuffix.docs.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+## When to use it
+
+- When there is a commonly understood symbol or abbreviation for the type of information we are asking for.
+- When the user needs to enter a value which has specific units.
```
edufarre, line 29:
```suggestion
- When the user needs to enter a value that has a specific unit.
```

docs/components/MaTextFieldWithSuffix.docs.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+## Variants
+
+[https://www.figma.com/embed?embed_host=notion&url=https%3A%2F%2Fwww.figma.com%2Ffile%2F7RhweKibaUnxX6WYBl7zKW%2FDesign-System-Beta%3Fnode-id%3D2572%253A1707](https://www.figma.com/embed?embed_host=notion&url=https%3A%2F%2Fwww.figma.com%2Ffile%2F7RhweKibaUnxX6WYBl7zKW%2FDesign-System-Beta%3Fnode-id%3D2572%253A1707)
+
+## Behavior
+
+Same as text-field.
+
+## References
```
edufarre, line 45: Not sure this is needed in our components library documentation. However, I believe it would be nice to have it referenced in our Figma design.

src/components/MaTextField/MaTextField.stories.js
```diff
@@ -110,17 +111,20 @@ TextFieldWithIcon.argTypes = {
...
   ...hideComponentProperties,
   suffix: {
     name: 'Suffix',
     defaultValue: '€',
     control: { type: 'text' },
   },
 }
+TextFieldWithSuffix.parameters = {
+  docs: { page: docsMaTextFieldWithSuffix },
+}
```
edufarre, line 130: Why is this set here?
```suggestion
```

I do believe it would be better if we keep consistency across our different stories implementation. This being said, documentation should be added in the `export default`, thus being something like:
```js
export default {
  title: 'Components/TextFIeld',
  // ...
  parameters: {
    docs: { page: docs },
  },
}
```

In a nutshell, I believe we need a 'standard' documentation for our `TextField` and then we could have more specific documentation in each of the examples or have all the different use cases documented in a unique doc file

docs/components/MaTextFieldWithSuffix.docs.mdx
cillascillas: I'd rather say `Input field with a suffix` and that's all. It is visible as it might be expected, isn't it?

docs/components/MaTextFieldWithSuffix.docs.mdx
cillascillas: This documentation was written by the Design team so I understand that they want this to be part of the docs too.

src/components/MaTextField/MaTextField.stories.js
cillascillas: The docs for TextField are already set as you mention (in the `export default`). However, correct me if I misunderstood it, you suggest having a unique file of docs for all kinds of TextFields (and not a page for each of them). If that's it, should we move this content to [TextField docs file](https://github.com/holaluz/margarita/blob/master/docs/components/MaTextField.docs.mdx)?

docs/components/MaTextFieldWithSuffix.docs.mdx
edufarre: If they think that this should be here, fine by me! 🚀  I just found that a bit odd.

src/components/MaTextField/MaTextField.stories.js
edufarre: To be honest, I do think we should add MaTextField generic documentation first. Having a single page with the entire documentation depends on how long and specific we would like each of the documented sections to be.

Actually, I'd be fine with either of the previously commented approaches  😄

docs/components/MaTextFieldWithSuffix.docs.mdx
edufarre: You're right, straight to the point, there is no need in saying so 💯

edufarre (approved):

alexcastelo95 (approved):

src/components/MaTextField/MaTextField.stories.js
cillascillas: That's a good idea, and in fact, we should discuss it or bring it to de DS meeting so everybody is in the same page, don't you think? (Then we could reuse this info and add it to the single page with the full doc for MaTextField)

edufarre: Yes, yes and yes! 🚀

PR #383, opened 2021-03-24
fix: Make MaTextField fluid

### LINKED ISSUE #382: weird behaviours for MaLayout for narrow containers
Weird stuff is going on with MaLayout for narrow containers, when some of the columns are too small. Sometimes it overflows, leaving part of the content outside the grid:

image

In this specific example, the ma-text-fields on the last row seem to be larger than they should and they break the whole layout.

I have two main suspicions:
 - There is css in ma-text-field that makes ma-text-field not fluid. For example some pesky min-width or something like that that breaks the column sizing. We have searched and it doesn't seem to be the case, but it could be something like that.
 - There is a problem with the function that calculates column sizes and it breaks for small widths.

afontcu:
Closes #382 

From:

Captura de Pantalla 2021-03-24 a les 16 04 44

To:

Captura de Pantalla 2021-03-24 a les 16 04 35

---

Now, this could be seen as a breaking change because some layouts might change. I'd say, though, that the expectation is that MaTextField would take all available space, so not having this behavior was a bug.

Thoughts?

### DIFF
--- a/src/components/MaTextField/MaTextField.css
+++ b/src/components/MaTextField/MaTextField.css
@@ -45,7 +45,7 @@
   flex: 0 1 100%;
   transition: border 0.2s ease;
   padding: 0 0.5rem;
-  max-width: 100%;
+  width: 100%;
   height: 100%;
   font-size: 1rem;
   border: none;


edufarre:
> Now, this could be seen as a breaking change because some layouts might change. I'd say, though, that the expectation is that MaTextField would take all available space, so not having this behavior was a bug.

Due to this was the 'expected' behaviour I wouldn't consider it a breaking change. If some templates do change, we must fix them in order to have consistency across the different projects when using this component.

edufarre (approved):

Samitier (approved):
oh shit this will be tough to debug. But I think its necessary. Good job!

sergigraciaj (approved):

PR #384, opened 2021-03-26
fix: Remove padding from MaOption

afontcu:
UI components should not interfere with external positioning/laying out – it is on Layout components to handle it.

This PR removes the additional padding surrounding every `MaOption`, which breaks the DS-based designs when using them.

Bear in mind that this might change some layouts.


edufarre (approved):
🚀

alexcastelo95 (approved):
LGTM! 🔥

cillascillas (approved):

PR #385, opened 2021-03-29
fix: Fix negative line-height crop

### LINKED ISSUE #379: MaText does not handle vertical cropping properly
Given the following snippet:

```vue
<ma-text>I'M a text</ma-text>
<ma-text>I'M a text</ma-text>
```

this is the current result:

Captura de Pantalla 2021-03-22 a les 10 32 35

whereas this is what we should expect:

image

afontcu:
Closes #379

Notice that it won't look perfect on every screen/SO/browser/config, but I feel this is closer to what we had in mind.

Captura de Pantalla 2021-03-29 a les 11 57 40
Captura de Pantalla 2021-03-29 a les 11 57 50
Captura de Pantalla 2021-03-29 a les 12 02 48
Captura de Pantalla 2021-03-29 a les 12 02 54


edufarre (approved):
Wicked 🚀 

I've seen we do have more accurate margin values, good job!

afontcu:
This is not ready to merge because there are some issues with MaList alignment

afontcu:
Ok, so now I think MaList works properly with MaText-based lists. I had to revamp it a bit and render the marker element (instead of relying on CSS and `<ol>`/`<ul>` elements. Also, I needed to add a couple of additional (but backwards-compatible!) props in Stack and Icon. 

Please take a look and see if I'm missing anything 🙏 @holaluz/frontend

src/components/MaIcon/MaIcon.vue
```diff
@@ -76,6 +76,14 @@ export default {
       default: 18,
     },
 
+    /**
+     * Sets icon's view box width
```
edufarre, line 80:
```suggestion
     * Sets icon's view box height
```

src/components/MaList/MaList.spec.js
```diff
@@ -6,72 +6,56 @@ describe('List', () => {
...
-    expect(getByRole('list').nodeName).toBe('OL')
+    expect(firstListItem).toHaveClass('ma-list-item')
+    expect(firstListItem).toContainElement(firstItem)
   })
 
-  test(`'size' property assigns the valid font size to list item`, () => {
-    const { getByText } = renderComponent({
-      size: 'medium',
-    })
+  describe('type', () => {
```
edufarre, line 16:
```suggestion
  describe(`type component's property`, () => {
```

src/components/MaStack/MaStack.vue
```diff
@@ -59,6 +59,16 @@ export default {
       default: 'fill',
       validator: responsivePropValidator(Object.keys(alignment)),
     },
+
+    /**
+     * Set the HTML element tag that wraps the content.
+     * @values div, ul, ol
+     */
+    tag: {
```
edufarre, line 67: Shall we add a test for this component property? 😇

src/components/MaList/MaList.vue
```diff
@@ -15,87 +9,49 @@ import { text, tones } from '../../tokens'
...
+      // tests complain about a non-vue instance being created and removed.
+      // Thus, we simply set an empty div.
+      return createElement('div')
+    }
+
+    const tag = props.type === 'ordered' ? 'ol' : 'ul'
+
+    return createElement(
+      'ma-stack',
+      { props: { space: 'small', tag } },
```
edufarre, line 41: Will the rendered stack always have the same gap spacing? (`small`)
I believe we could add a prop with `default: 'small'` for this, thus wrapping our stack element into a `ma-list` component

edufarre (approved):
Overall, great job! 🔥

src/components/MaList/MaList.vue
afontcu: > Will the rendered stack always have the same gap spacing? (small)

It will have the same spacing as it has before – this PR intended to be non-breaking for Lists, so if we need to add some spacing props on List, we could always add it afterwards 👍

src/components/MaStack/MaStack.vue
afontcu: sure, done!

PR #386, opened 2021-03-29
feat: add bold and tone props for label in MaTextField

sergigraciaj:
As proposed in the design for the new solar product

image


alexcastelo95 (approved):
Looks Good Sergi 👍🏽

afontcu:
Given this and the fact that @Samitier mentioned the need to render a tone="white" for a MaOption, what if we simply allowed all our Form elements to accept tone/weight props, just like MaText/Heading does, so we kill all these requirements at once? Needing to pass extra props to pretend we don't need to style the inner MaText of these components feels a bit unnatural

src/components/MaTextField/MaTextField.vue
```diff
@@ -7,6 +7,8 @@
         size="small"
         tag="label"
         class="ma-text-field__label"
+        :tone="labelTone"
+        :bold="isLabelBold"
```
afontcu, line 11: Looks like props got renamed but not these ones (smells like a missing test 😉)

sergigraciaj: 🙈 
thx!

afontcu (approved):
LGTM – make sure you update the title accordingly!

Issue #387, opened 2021-03-30
MaLayout sets the vertical spacing for hidden elements


Samitier:
If we have something like this:

```html
<ma-layout columns="12 - 12 - 12" gap="medium">
    <div>Something</div>
    <div>Something</div>
    <div v-if="false">Something hidden</div>
</ma-layout>
```

It will still render the gap for the hidden row:
![image](https://user-images.githubusercontent.com/4160121/112962688-f17a6280-9146-11eb-9f3f-5a82f1e529c8.png)

We should check if there are children in those elements before rendering them in `MaLayout`.

afontcu:
Since this is not happening with MaStack, am I correct to assume that the issue must be on Layout's `renderGrid`? https://github.com/holaluz/margarita/blob/master/src/components/MaLayout/MaLayout.vue#L120

Samitier:
> Since this is not happening with MaStack, am I correct to assume that the issue must be on Layout's `renderGrid`? https://github.com/holaluz/margarita/blob/master/src/components/MaLayout/MaLayout.vue#L120

Yes! I think we should filter out empty `rowChildren`:
https://github.com/holaluz/margarita/blob/ec3c8ed5510eeafc1c520df6e6d5baa762c931e5/src/components/MaLayout/MaLayout.vue#L127

PR #389, opened 2021-03-30
fix: set correct vertical gap for hidden elements

Samitier:
Fixes: https://github.com/holaluz/margarita/issues/387


src/components/MaLayout/MaLayout.vue
```diff
@@ -120,25 +120,27 @@ function renderStack({ createElement, dom, gap, data }) {
...
+        createElement,
+        columns: row.join(' '),
+        children: rowChildren,
+        gap,
+        justify,
+        verticalAlign,
+      })
     })
-  })
+    .filter((a) => a)
```
Samitier, line 143: do you prefer if I do a for loop instead of this?

afontcu: Hm, is this really needed even with the `if (!rowChildren.length) return` statement? Looks like it should be enough.

If not, `filter(Boolean)` would work, too, but I'm not sure which one is more magical XD

src/components/MaLayout/MaLayout.spec.js
```diff
@@ -43,11 +63,10 @@ describe('MaLayout', () => {
   })
 })
 
-function renderComponent(props, attrs = {}) {
+function renderComponent(props, slots) {
   const utils = render(MaLayout, {
     props: { gap: 'small', columns: '12', space: 'small', ...props },
-    slots: { default: 'content' },
-    ...attrs,
+    slots: { default: slots || 'content' },
```
edufarre, line 69: What about?
```js
function renderComponent(props = {} , slots = 'content') {
  const utils = render(MaLayout, {
    props: { gap: 'small', columns: '12', space: 'small', ...props },
    slots: { default: slots },
})
```

afontcu (approved):

edufarre (approved):

PR #391, opened 2021-04-01
chore: do not load legacy fonts

Samitier:
Let's not have fonts for browsers that we do not support.

image
Source: https://medium.com/@aitareydesign/understanding-of-font-formats-ttf-otf-woff-eot-svg-e55e00a1ef2


alexcastelo95 (approved):
🧨

edufarre (approved):
image

carlotamolintr (approved):

afontcu (approved):
neat 👌  we should remove these files from the S3 bucket, too

PR #393, opened 2021-04-06
feat: add MaContainer

### LINKED ISSUE #375: [RFC] MaContainer component
Hello team,

There is one issue is been bugging me for a while and that is the lack of a "container" layout component. I honestly think it should be in our DS so here's my idea.

## What is a container?

I'm refering to a "container" (in lack of a better name) to something that has a responsive max-width a margin: auto and a padding. For example, this:

image

In this specific example (you can see it [here](https:///alta-staging.holaluz.com) in action), this container has a variable max-width, depending on the width of the screen. It also has a margin auto to always center content in the middle and a padding, so that mobile layouts don't "glue" to the sides of the screen.

## Why variable max-width? Why not a single max-width?

This is a legacy issue that the DS should try to fix IMHO. This is a snippet of what our old DS did:

```css
@include mq($from: sm) {
    max-width: 540px;
}

@include mq($from: md) {
    max-width: 720px;
}

@include mq($from: lg) {
    max-width: 960px;
}

@include mq($from: xl) {
    max-width: 1180px;
}
```
Changing this will break designs in dryMartini and maitai. I'm just saying this so that we have it in mind for this RFC or even for deciding if we need this component in the DS or not.

I also kinda like the idea of having multiple max-widths, even if it means losing "screen real estate" for smaller devices. It's easier to do the responsive if we have the max-widths more controlled.

Of course, if we do "responsive" max widths, they should be done with the responsive props that we have right now in margarita (not with the mess that we had before).

Well, this is something we should discuss together in the DS meetings anyways.

## API example:

The only prop we should have is the `max-width`. Which should depend on what the DS sets as our standard container sizes. If there is the same max-width on all of our apps (which I doubt), remove the prop altogether.
```vue
<ma-container
    size="medium" // the standard container size, for example maitai.
/>
```
```vue
<ma-container
    size="small" // a small container size, for example embeds or email templates
    // (if we could put margarita on emails hahah)
/>
```
```vue
<ma-container
    size="large" // large container sizes, for example dashboards
/>
```
We could also explore the idea of a variable size using the responsive props (passing an array like ma-stack space).

What do you guys think? Would this be useful? Is it too situational and we will only use it in dry/maitai/oneclicksales? 

Let's discuss it! 😃

afontcu:
Closes #375

### DIFF
--- a/src/components/MaContainer/MaContainer.spec.js
+++ b/src/components/MaContainer/MaContainer.spec.js
@@ -1,0 +1,13 @@
+import { render } from '@margarita/margarita-test-utils'
+import MaContainer from './MaContainer'
+
+describe('Container', () => {
+  test('renders a container', () => {
+    const { getByText } = render(MaContainer, {
+      slots: { default: 'Hello World' },
+      props: { size: 'small' },
+    })
+
+    expect(getByText(/hello world/i)).toHaveClass('ma-container--small')
+  })
+})
--- a/src/components/MaContainer/MaContainer.stories.js
+++ b/src/components/MaContainer/MaContainer.stories.js
@@ -1,0 +1,25 @@
+import MaContainer from '../MaContainer'
+
+export default {
+  title: 'Components/Container',
+  component: MaContainer,
+  argTypes: {
+    size: {
+      control: {
+        type: 'select',
+        options: ['small', 'medium', 'large'],
+      },
+    },
+  },
+}
+
+const Template = (args, { argTypes }) => ({
+  props: Object.keys(argTypes),
+  template: `<ma-container v-bind="$props" style="background-color: var(--color-pink-base)">
+    <ma-text tone="white">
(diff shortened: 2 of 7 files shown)


alexcastelo95 (approved):
Just curious. Do we want to have the container variables only in the component or does it make sense to have them as DS tokens in the [tokens](https://github.com/holaluz/margarita/tree/add-container/src/tokens) folder?

Very nice work. 🔥

afontcu:
> Do we want to have the container variables only in the component or does it make sense to have them as DS tokens in the [tokens](https://github.com/holaluz/margarita/tree/add-container/src/tokens) folder

I left them here because they are only used in MaContainer (as of today), and because I'm not sure we want to expose them as tokens (would you need to use `var(--container-small)` somewhere in the project? I don't think so). 

That being said, I'm completely open to discuss!

alexcastelo95:
> > Do we want to have the container variables only in the component or does it make sense to have them as DS tokens in the [tokens](https://github.com/holaluz/margarita/tree/add-container/src/tokens) folder
> 
> I left them here because they are only used in MaContainer (as of today), and because I'm not sure we want to expose them as tokens (would you need to use `var(--container-small)` somewhere in the project? I don't think so).
> 
> That being said, I'm completely open to discuss!

Nothing, it was just to know the criteria on the basis of which you decided. I think it makes sense to leave it on the component since we don't have any other uses.

Thank you for your response. 😃

PR #394, opened 2021-04-09
feat: Add size and tone props to ma-option

carlotamolintr:
These two props solve the text alignment issue in call me component. 
image

Cause for an specific size text will be a margin-top to align the text with the checkbox.

Current issue in kali:

We are duplicating the ma-text component in call me, the first one is being used by the ma-option and the second one is being used to apply size and tone. The second one is making issues with the positioning so with this PR we could remove one

### DIFF
--- a/src/components/MaOption/MaOptionCheckbox.css
+++ b/src/components/MaOption/MaOptionCheckbox.css
@@ -18,20 +18,14 @@
   padding-left: calc(var(--indicator-size) + 0.5rem);
   color: var(--color-gray-dark);
 }
-.ma-selector-xsmall {
-  &.ma-option--checkbox .description {
-    margin-top: 4px;
-  }
+.ma-selector-xsmall.ma-option--checkbox .description {
+  margin-top: 4px;
 }
-.ma-selector-small {
-  &.ma-option--checkbox .description {
-    margin-top: 3px;
-  }
+.ma-selector-small.ma-option--checkbox .description {
+  margin-top: 3px;
 }
-.ma-selector-medium {
-  &.ma-option--checkbox .description {
-    margin-top: 2px;
-  }
+.ma-selector-medium.ma-option--checkbox .description {
+  margin-top: 2px;
 }
 
 .ma-option--checkbox .input {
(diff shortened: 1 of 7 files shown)


afontcu (approved):
Looks good to me! 🚀

edufarre (approved):

src/components/MaOption/MaOptionCheckbox.css
```diff
@@ -16,9 +16,23 @@
...
+.ma-selector-small {
+  &.ma-option--checkbox .description {
+    margin-top: 3px;
+  }
+}
+.ma-selector-medium {
+  &.ma-option--checkbox .description {
+    margin-top: 2px;
+  }
+}
```
Samitier, line 35: this nesting is useless no? And `ma-selector-xxx` is always in `ma-option--checkbox` so why not:
```suggestion
.ma-selector-xsmall .description {
  margin-top: 4px;
}
.ma-selector-small .description {
  margin-top: 3px;
}
.ma-selector-medium  .description {
  margin-top: 2px;
}
```

src/components/MaOption/MaOptionRadio.css
```diff
@@ -10,9 +10,23 @@
...
+.ma-selector-small {
+  &.ma-option--radio .description {
+    margin-top: 3px;
+  }
+}
+.ma-selector-medium {
+  &.ma-option--radio .description {
+    margin-top: 2px;
+  }
+}
```
Samitier, line 27: same, useless nesting & redundant `ma-option--radio` selector, no?

Samitier (approved):
just a tiny stuff with the css selectors and lgtm!

src/components/MaOption/MaOptionRadio.css
carlotamolintr: nesting removed!

src/components/MaOption/MaOptionCheckbox.css
carlotamolintr: nesting removed!

PR #399, opened 2021-04-15
fix: error in ma-text-field

carlotamolintr:


src/components/MaTextField/MaTextField.vue
```diff
@@ -14,12 +14,16 @@
       <!-- @slot Label's sibling content slot -->
       <slot name="labelSibling" />
     </div>
-    <div class="ma-text-field__input-wrapper" :class="inputWrapperClasses">
+    <div
+      class="ma-text-field__input-wrapper"
+      :class="suffix ? inputWrapperClasses : ''"
```
edufarre, line 19:
```suggestion
      :class="inputWrapperClasses"
```
⚠️  It continues on the other comment

src/components/MaTextField/MaTextField.vue
```diff
@@ -143,11 +147,15 @@ export default {
     inputWrapperClasses() {
       return {
         'ma-text-field__input-wrapper--disabled': this.$attrs.disabled,
-        'ma-text-field__input--error': this.hasError,
-        'ma-text-field__input--error-icon': this.hasError && !this.suffix,
+        'ma-text-field__input--error': this.hasError && this.suffix,
       }
```
edufarre, line 151:
```js
return this.suffix ? {
        'ma-text-field__input-wrapper--disabled': this.$attrs.disabled,
        'ma-text-field__input--error': this.hasError,
      } : ''
```
or
```js
return {
        'ma-text-field__input-wrapper--disabled': this.$attrs.disabled && this.suffix,
        'ma-text-field__input--error': this.hasError && this.suffix,
      }
```

carlotamolintr: Hi @edufarre,
Thank you for your feedback. I make it as you told me and also I had to added an extra condition (cause I was forgetting an style). COuld you please check again please?

Thnx :)

src/components/MaTextField/MaTextField.stories.js
```diff
@@ -90,7 +90,6 @@ TextFieldWithSiblingLabel.argTypes = {
     defaultValue: 'Additional label',
     control: { type: 'text' },
   },
-  ...hideComponentProperties,
```
afontcu, line 93: Looks like `hideComponentProperties` is not being used anymore? Is it accurate? If so, can it be removed?

src/components/MaTextField/MaTextField.vue
```diff
@@ -141,13 +142,21 @@ export default {
...
 
     inputWrapperClasses() {
+      return this.suffix
+        ? {
+            'ma-text-field__input-wrapper--disabled': this.$attrs.disabled,
+            'ma-text-field__input--error': this.hasError && this.suffix,
+          }
+        : !this.suffix && this.hasError
+        ? 'ma-text-field__input--error'
+        : ''
```
afontcu, line 152: 💡 Could we avoid nested ternary statements? I'd say these are quite hard to read and reason about. Also, some stuff is a bit redundant – why would we need to check for `this.suffix` on line 148 if we already know it is true?

I'd argue that having several if statement + early returns would yield a more readable code.

That being said, though, this is purely a personal opinion, so feel free to ignore it :D

src/components/MaTextField/MaTextField.vue
```diff
@@ -141,13 +142,21 @@ export default {
...
+    },
+
+    inputClasses() {
       return {
-        'ma-text-field__input-wrapper--disabled': this.$attrs.disabled,
-        'ma-text-field__input--error': this.hasError,
-        'ma-text-field__input--error-icon': this.hasError && !this.suffix,
+        ' ma-text-field__input--error-icon': this.hasError && !this.suffix,
```
afontcu, line 157:
```suggestion
        'ma-text-field__input--error-icon': this.hasError && !this.suffix,
```

src/components/MaTextField/MaTextField.spec.js
```diff
@@ -22,8 +22,21 @@ describe('TextField', () => {
...
     })
 
-    expect(input.parentNode).toHaveClass('ma-text-field__input--error')
-    expect(input.parentNode).toHaveClass('ma-text-field__input--error-icon')
+    expect(input.parentNode).toHaveClass('ma-text-field__input-wrapper')
+
+    expect(getByText(errorMessage)).toBeInTheDocument()
+  })
+
+  test('renders error CSS class', () => {
```
afontcu, line 30: Should we add a test for the suffix+error case?  Or is it already covered?

afontcu (commented):
I've added a few comments here and there! Looks good overall, though :) 

💡 maybe adding an additional story with several prop configurations would help us avoid regressions – Percy would catch any undesired change

src/components/MaTextField/MaTextField.stories.js
carlotamolintr: removed!

src/components/MaTextField/MaTextField.vue
carlotamolintr: thanks for help to clean 🤣

src/components/MaTextField/MaTextField.spec.js
carlotamolintr: done

afontcu (approved):

edufarre (approved):

PR #401, opened 2021-04-16
feat: Update alert styling and remove size prop

cillascillas:
After a [design review, new updates related to styling](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/Design-System?node-id=5314%3A1401) are coming:

- Title font-size -> 20px
- Body font-size -> 16px
- Link styling (depending on alert type, underlined by default, not underlined in hover)
- Remove alert size prop:
    - Type prop would be removed.
    - Default size would be current medium. (It only affects to padding size)
- Add new doc

### DIFF
--- a/docs/components/MaAlert.docs.mdx
+++ b/docs/components/MaAlert.docs.mdx
@@ -1,3 +1,68 @@
 # Alert
 
-This is the documentation section about Alert component
+## Description
+
+Alerts provide relevant information to the user without impeding their task. Alerts are system generated and may or may not result from a user’s action.
+
+## How to use it?
+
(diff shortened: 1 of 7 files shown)


afontcu:
We could also remove the `size` prop, right? 🔥🔥🔥 🚒

cillascillas:
@afontcu 🤦‍♀️ sure, my bad! That was the reason for the PR in the first place!

afontcu (approved):
🚀 🔥

alexcastelo95 (approved):
🚀

PR #402, opened 2021-04-19
Add prevent close on modal

sergigraciaj:
As we can have asynchronous events initiated in the modal, this PR aims to prevent closing the modal via prop.

### DIFF
--- a/src/components/MaModal/MaModal.stories.js
+++ b/src/components/MaModal/MaModal.stories.js
@@ -20,11 +20,6 @@ export default {
         options: ['white', 'gradient'],
       },
     },
-    preventClose: {
-      control: {
-        type: 'boolean',
-      },
-    },
   },
   parameters: {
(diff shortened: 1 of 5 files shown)


Samitier:
Is it not better to hide the close button? Why do we want a close button that does nothing?

src/components/MaModal/MaModal.stories.js
```diff
@@ -20,6 +20,11 @@ export default {
         options: ['white', 'gradient'],
       },
     },
+    preventClose: {
```
edufarre, line 23: Is this really need it? I guess storybook controls already know it should be a boolean control by looking into the property type.

sergigraciaj: Thx 👍

sergigraciaj:
As we discussed in the Design System meeting, we don't want this behaviour in the modal.

PR #406, opened 2021-04-29
fix: Add gap support for older browsers

### LINKED ISSUE #388: Make MaLayout/MaStack/MaColumn compatible with MacOS 11x
Just a tiny bug I found with browserstack: `gap` doesn't work. We need to use `grid-gap`:

image

Seems like a very easy fix. Is it worth it?

afontcu:
Closes #388.

With this change, postcss is able to detect the `gap` property and add the appropriate fallback for shitty browsers:

imatge
_(screenshot from dist/margarita.css)_


Samitier (approved):
thank you ❤️ ❤️

Samitier:
> the appropriate fallback for shitty browsers
😂

edufarre (approved):

PR #407, opened 2021-04-30
docs: add figma embed color to docs

carlotamolintr:
This PR is to test the figma embeds in our docs.

You can se the example in the **Introduction category** --> **Color System** (you will need to scroll down)

Process:

To get the correct link we need to group the elements we want to show in FIGMA. Otherwise, the embed will show all the page.

Will be interesting to  confirm that url created in FIGMA will remain exactly the same after the Figma release

Please share your opinions :)

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -12,7 +12,7 @@ export const parameters = {
     storySort: {
       order: [
         'Introduction',
-        ['Welcome', 'Getting Started', 'Design System'],
+        ['Welcome', 'Getting Started', 'Design Principles', 'Color System'],
         'Tokens',
         'Layout',
         'Components',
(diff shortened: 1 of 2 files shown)


carlotamolintr:
At least, we tried embeding all the (colors) figma file.  Doing this, the total size for the storybook is ~18MB 😅.
So, if we want to show an embed for each section docs, the storybooks will be heavier.

sergigraciaj:
We considered using addons like [this one](https://storybook.js.org/addons/storybook-addon-designs) or [this](https://storybook.js.org/addons/storybook-figma)?

carlotamolintr:
A agreed yesterday, I will merge this PR and we keep the figma addons links to investigate later in another PR. 
Thank you all

sergigraciaj (approved):

alexcastelo95 (approved):

PR #408, opened 2021-05-03
feat: MaTabs component

afontcu:
Implements a Tabs component that follows the Design System.

- With icon and pills:

image

- Or simple, with only a title:

image


src/components/MaTabs/MaTabs.vue
```diff
@@ -0,0 +1,94 @@
...
+    const tabs = this.$slots.default.filter((c) => c.tag)
+
+    const tabList = []
+    tabs.forEach((child, index) => {
+      const { title, icon, pill } = child.componentOptions.propsData
+
+      const tabHeaderContent =
+        this.variant === 'underline' ? (
+          <ma-layout gap="small" columns="*" verticalAlign="center">
+            {icon && <ma-icon icon={icon} />}
```
carlotamolintr, line 41: I don't understand that... if there is a prop icon on the tabs what is happening in this line? will come a string and the icon?🧐

afontcu: We're making sure we only render a `<ma-icon>` if `icon` is set. The line [uses the && operand to short-circuit](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#short-circuit_evaluation), so it is the same as writing

```js
if (icon) {
  <ma-icon icon={icon} />
}
```

In short: it works because `true && expression` evaluates to expression, and `false && expression` evaluates to false.



Hope this helped!

carlotamolintr: so, is a v-if condition to be sure that if there is not icon, will print nothing avoiding the error?

afontcu: Yes! It is "just javascript":

`true && console.log('hola!')` logs "hola!"

`false && console.log('hola!')` logs nothing

src/components/MaTabs/MaTabs.vue
```diff
@@ -0,0 +1,94 @@
...
+            <ma-text bold>{title}</ma-text>
+            {pill && <ma-pill text={pill} />}
+          </ma-layout>
+        ) : (
+          <ma-layout gap="small" columns="12 - 12">
+            <ma-text bold>{title}</ma-text>
+            <div>{child.data.scopedSlots?.header()}</div>
```
carlotamolintr, line 48: This child.data.scopedSlots is the scopedSlots present in tabs, right? But I don't know the ?.header where are you taking from... Could you explain me😓?

afontcu: sure! 

We want to get the `<template #header>hola</template>` named slot from MaTab, and render it on the header tab.

So:
`child` is the component itself, the Tab (we got it from the forEach loop).
`data.scopedSlots` let us access named and scoped slots. This can be undefined, though, and this is why I added `?`, the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).
`header()` is the named slot we want to access. It is a function because… Vue says so ([here](https://vuejs.org/v2/api/#vm-scopedSlots)) 😂

alexcastelo95 (commented):
I've reviewed and I think I follow the code, although I've picked up some Vue concepts I wasn't familiar with.

I'm not confident enough to give the approval as there are certain parts that escape me and I'd rather let other colleagues see it.

Very good work.

src/components/MaTabs/MaTabs.vue
sergigraciaj: Why don't we have the header as a prop in Ma-Tab?

afontcu: Because you don't want to set a different header for each tab, right? But instead for the whole Tabs block

sergigraciaj: By header we mean this, right?
image
It's different in every tab, isn't it?

afontcu: ah, I thought you meant the underline/gradient header type. I see.

How would you pass that as a prop? Only way I see (other than a slot) is by passing a component, which isn't something Vue supports.

cillascillas:
👋  I like the approach! 
(And also that you've ended using JSX because it makes it clearer to read/understand for a component of this size)

I was wondering if maybe adding the tab content wrapped up in a `<template #content>`  would make it clearer or not (instead of the element right away) and I don't know if it would complicate the implementation (if it's a yes then, disregard). But apart from that, I like the idea of all the things related to the tab together in the same <ma-tab>, so nice! 🙂 

As you stated in the description apart from `title` (which is mandatory for both) pill and icon are not and are not used in gradient tabs so maybe adding a warn if we use them incorrectly. Definitely not a must because we shouldn't do it, but just thinking out loud.

So, for me that's a **+1** for moving on with the rest of the implementation following this approach!!

src/components/MaTabs/MaTabs.vue
sergigraciaj: Ok, I get it, it's because it's not just a text, we can have a composed text with differents sizes and weigths.
I know this PR is not for styling, but is the tab responsability to set the color to the headder/composed text then, right?

afontcu: to be honest, I'd leave the #header slot content styling to the consumer. 

Other than that, yeah, I think `title` styling should be defined from the component 👍

sergigraciaj: I'm ok with that, good work ^^

edufarre (commented):
First of all, such a nice approach. Even though you used JSX, it's quite clear which elements/template are being rendered in each case, the code is properly encapsulated and easy to follow what happens, good job!

I would like to say that I agree with @cillascillas, although it might be a bit redundant I believe that having the 'tab content' inside something like `<template #content>Whatever </template>` would make the skeleton easier to read/understand.

Overall, I like this approach and it looks like the way to go for me, the API is not complicated and the component's body is quite simple to understand.

src/components/MaTabs/MaTabs.vue
edufarre: Agree. I believe we should be aware enough to not do really odd things with the `#header`'s style

src/components/MaTabs/MaTabs.vue
```diff
@@ -0,0 +1,96 @@
...
+  },
+}
+</script>
+
+<style lang="postcss" scoped>
+ul {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
```
Samitier, line 91: a bit offtopic but we have this same 3 lines of css repeated 34 times in dry-martini (we don't want to add it as global styles in dry-martini to not break any component from margarita). 

Can we consider setting this as global css here in margarita? Do we use vanilla HTML lists anywhere?

afontcu: Sure, sanitize.css is not providing these sort of resets, but we could consider adding it somehow 👍

afontcu:
> How could you use them wrong?

You could do something like:

```vue
<ma-tabs variant="gradient">
  <ma-tab icon="user" pill="12" title="hey">...</ma-tab>
</ma-tabs>
```

...and `icon` and `pill` would have no effect (because we're rendering a gradient tabs). We could check that when `mount`ing a tab!

edufarre:
> > How could you use them wrong?
> 
> You could do something like:
> 
> ```
> <ma-tabs variant="gradient">
>   <ma-tab icon="user" pill="12" title="hey">...</ma-tab>
> </ma-tabs>
> ```
> 
> ...and `icon` and `pill` would have no effect (because we're rendering a gradient tabs). We could check that when `mount`ing a tab!

Gotcha! Then I believe having a check as you suggested would be a good option! I understood that like it was related to the usage of the actual pill or icon per se, my bad. Thanks for the example 👍🏽

Samitier:
Amazing work bro! Some feedback after playing with these tabs a bit:

## Accessibility stuff
Since it's a PoC I guess its too soon to implement this, but I think the tabs should be focusable with tab (the keyboard tab 😂 ). Maybe we need a "focused" style too.

## API nitpicking number one
Ok so I wanted to show a custom header only if the current tab was active. Putting v-if in the template header breaks the code:

```vue
<ma-tab title="title 1">
  <template v-if="selectedTab === 0" #header>With a custom header</template>
  <ma-text>Hello from tab number 1!</ma-text>
</ma-tab>
<ma-tab title="title 2">
  <template v-if="selectedTab === 1" #header>With a custom header</template>
  <ma-text>Hello from tab number 2!</ma-text>
</ma-tab>
<ma-tab title="title 3">
  <template v-if="selectedTab === 2" #header>With a custom header</template>
  <ma-text>Hello from tab number 3!</ma-text>
</ma-tab>
```
image

Also to check the selected tab I have to put magic numbers in the code. I kinda don't like that. The second point is related to this:

## API nitpicking number two

Ok so, as I said I don't like that the selected tab is an integer. Why? Because I think I will never render tabs without a `v-for`:

This is "too manual"

```vue
 <ma-tabs v-bind="$props" v-model="selectedTab" @input="input" variant="gradient">
    <ma-tab title="title 1">
      <template #header>With a custom header</template>
      <ma-text>Hello from tab number 1!</ma-text>
    </ma-tab>
    <ma-tab title="title 2">
      <template #header>With a custom header</template>
      <ma-text>Hello from tab number 2!</ma-text>
    </ma-tab>
    <ma-tab title="title 3">
      <template #header>With a custom header</template>
      <ma-text>Hello from tab number 3!</ma-text>
    </ma-tab>
</ma-tabs>
```

It's more reasonable to think I would do this:

```vue
 <ma-tabs v-bind="$props" v-model="selectedTab" @input="input" variant="gradient">
    <ma-tab v-for="i in 3" :key="i" title="title {{i}}">
      <template #header>With a custom header</template>
      <ma-text>Hello from tab number {{i}}!</ma-text>
    </ma-tab>
</ma-tabs>
```

And therefore, I would like my API to render tabs based on an array and do the `v-for` for me. For example:

```vue
 <ma-tabs :tabs="['foo', 'bar']">
    <template #header="{tab, isSelected}">
        this is the header for {{ tab }}
        <div v-if="isSelected">And if it is selected, I show this</div>
    </template>
    <template #body="{tab}">
        this is the body for {{ tab }}
    </template>
</ma-tabs>
```

This would be great for copy management:

```vue
 <ma-tabs :tabs="$t('tabs')">
    <template #header="{tab}">
       {{ tab.title }}
    </template>
    <template #body="{tab}">
        {{ tab.description }}
    </template>
</ma-tabs>
// i18n yml
tabs:
  - title: foo
     description: description
  - title: bar
     description: description
```

But we could also do whatever we want:

```vue
<ma-tabs :tabs="[FooComponent, BarComponent]" v-model="selectedTab">
    <template #header="{tab}">
      {{ tab === FooComponent ? 'foo' : 'bar' }}
    </template>
    <template #body="{tab}">
       <component is="tab" />
    </template>
</ma-tabs>
```

Also, the implementation of the scoped slots approach is substantially easier IMHO.

That's it. As I've said, its nitpicking after playing with the api a bit, nothing too important.

afontcu:
> I wanted to show a custom header only if the current tab was active

That snippet won't work because slot content is evaluated in child context, I think, so it won't know what the `v-if` values are.

Would this work?

```vue
<ma-tabs title="title 1">
  <my-cool-header #header :selectedTab="selectedTab" />
  <ma-text>Hello from tab number 1!</ma-text>
</ma-tab>
```

and then you can choose what to render depending on the prop value.

> Ok so, as I said I don't like that the selected tab is an integer. 

haha yeesss I thought nobody would complain about this! 😂 I don't like having a integer-based v-model, either.

I was thinking about using refs or something? With a scoped slot from MaTabs giving you the selected tab ref back?

> And therefore, I would like my API to render tabs based on an array and do the v-for for me

Let's bear in mind we're talking about tabs, which will be limited – we won't have like 15 tabs or whatever! All usages we've found consist of 2-3 tabs, not more. Thus, writing them "manually" instead of iterating over an artificial array doesn't seem like a bad tradeoff. Is it really that important to be able to use a `v-for`?

sergigraciaj:
I would stick to the proposed API as it give us freedom to compose the header and body with different components between tabs.

Samitier:
Hello!

I'm going to revive this PoC to see if we can advance with this. I'll add some comments:

## Change of scope of ma-tabs

We talked with the product design crew and we agreed in that one of the reasons this component is not released yet is that we were trying to solve too many existing problems at the same.

That's why we decided to re-focus its scope to solve a tinier problem and expand upon that. That example is the tabs we recently added in maitai in https://github.com/holaluz/maitai/pull/502

Since its a new feature, it's easier that we can base this PR in a real use case and release it. After that, incrementally adding new stuff to refactor old tabs should be easy and more "organic". Maybe we don't even need that the tabs are different between projects and a simpler approach will suffice for all of them.

## Accessible tabs

For instance, we decided that we don't need the "gradient tabs" because they are an accessibility problem and we will need to change them in the future. This simplifies this feature a lot.

## API proposal

Because of a tiny mistake (we completely forgot this PR existed) we already implemented a MaTab in  [the PR in maitai](https://github.com/holaluz/maitai/pull/502/files#diff-efe899b30388881f222c6f6fa7edae0f2aa13db5e628ce1d41136cbd7c15c1c4)

The API in there is a bit different than the one proposed here. It's implementation is also way simpler but less extendable because it does not allow a slot for the header.

So the proposed API in this PoC is this:

```vue
<ma-tabs v-bind="$props" v-model="selectedTab" @input="input">
    <ma-tab title="title 1">
       <ma-text>Hello from tab number 1!</ma-text>
    </ma-tab>
    <ma-tab title="title 2">
       <ma-text>Hello from tab number 2!</ma-text>
    </ma-tab>
</ma-tabs>
```

And the API in maitai is this:

```vue
<mai-tabs
   v-model="tab"
   :options="[
       { label: 'title 1', value: 'tab1' },
       { label: 'title 2', value: 'tab2' },
   ]"
>
  <template #tab1>
    <ma-text>Hello from tab number 1!</ma-text>
  </template>
  <template #tab2>
    <ma-text>Hello from tab number 2!</ma-text>
  </template>
</mai-tabs>
```

It would be amazing if we decided the API of this component together, so that I can open a PR with the feature. If no one has a preference or there are no answers and this thread dies I will go ahead and copy/paste the implementation of maitai and open a discussion there.

Thank you 😄

afontcu:
thanks for bringing this up!

Assuming that both API provide the functionalities we need:

I think it'd go with the API provided by this proof of concept, because it is closer to what HTML and Vue look like. It uses elements to render content, props to pass internal values… more natural. I'm worries about the "the selected tab is an integer" issue, though. It wasn't solved in the first iteration, even though we suggested some alternatives.

cillascillas:
Thanks Blai for reopening this! 
For me, the first option is clearer. I prefer having all together (The title and the content). In the second one, the drawback I find is that, given that you make the list of tabs in the object and then the templates for the content later, perhaps more errors can be done missing an item, not closing properly the tags or something like that... (I know that in the DS meeting we said that there shouldn't be no more than 7 tabs, so it's not such a big list though)

That said, it's just a matter of taste, so I'm not against the second one if we choose it eventually.

Samitier:
Time's up! Perfect, I'm going to do a PR over this PR to update it with what we said, keeping the same API.

Thank you both for your answer  ❤️

Samitier:
PR updated from what we talked avobe.

alex-castelo (approved):
Good job. Everything seems to be correct.
I only have doubts about the fact of declaring the props in the `MaTabs` component but I understand the justification of the comment. I wouldn't know how to do it.

Thanks for pushing for this. 💪🏽

Samitier (approved):

Issue #409, opened 2021-05-04
MaOption is broke when there is a image and text


carlotamolintr:
Hello all,

I guess after setting the tag="p" in ma-text as default the maOption is broken when you try to align img + text (please see the image below). I guess the fix is just to add the tag="span". But I would like to confirm.

![Captura de pantalla 2021-05-04 a las 13 01 26](https://user-images.githubusercontent.com/51370533/116994737-63f7d880-acd9-11eb-8fbd-bb7ea345929d.png)

<img width="738" alt="Captura de pantalla 2021-05-04 a las 13 03 15" src="https://user-images.githubusercontent.com/51370533/116994616-3ad74800-acd9-11eb-8d33-e000a3d80290.png">

Also you can check the issue here: https://deploy-preview-416--maitai.netlify.app/es/

Thank you all.

afontcu:
Yes, it make sense! Here and in other internal usages of MaText here in Margarita. Looks like a quick fix 😉

PR #410, opened 2021-05-04
fix: MaOption set span as default tag

carlotamolintr:
Related to https://github.com/holaluz/margarita/issues/409

Hello all,

I guess after setting the tag="p" in ma-text as default the maOption is broken when you try to align img + text (please see the image below). This pr is to fix it.

image

### DIFF
--- a/src/components/MaList/MaList.stories.js
+++ b/src/components/MaList/MaList.stories.js
@@ -12,18 +12,6 @@ export default {
         options: ['bullet', 'check', 'ordered'],
       },
     },
-    tone: {
-      control: {
-        type: 'select',
-        options: Object.keys(tones),
-      },
-    },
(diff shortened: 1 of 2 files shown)


sergigraciaj (approved):

afontcu (approved):
Looks good. However, are there other MaText usages inside Margarita that might need tweaking? It'd be great to check that.

carlotamolintr:
It looks like MaAlert is working being tag="p". I guess alll the other components that are using ma-text have already defined a tag.

carlotamolintr:
oh, no MaList doesn't have any default tag....

carlotamolintr:
At least, I removed tone and size controls from MaList. Also, I removed tag="p" in MaList

PR #411, opened 2021-05-04
feat(MaCard): Remove has-padding-top from

sergigraciaj:
As now we have Ma-Layout and Ma-Stack we don't need the `has-padding-top` anymore, so this PR removes the prop to prevent its usage.

It's a breaking change, but a "little" one, it's only used 9 times only in Bonasera.

I'm not sure how to tag this release, fix with breaking change in the comment to release a major version?

### DIFF
--- a/src/components/MaCard/MaCard.css
+++ b/src/components/MaCard/MaCard.css
@@ -10,7 +10,3 @@
   box-shadow: 0 0 6px 2px var(--shadow-light);
   background-color: var(--color-white-base);
 }
-
-.ma-card--has-padding-top {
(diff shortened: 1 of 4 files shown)


afontcu:
If the breaking change is only affecting Bonasera and you're willing to fix it right away after making the release, I'd go ahead with a `feat(MaCard): Remove has-padding-top from` release 👍

sergigraciaj:
Yup, of course : )

edufarre (approved):

cillascillas (approved):
Yes, please!

PR #412, opened 2021-05-04
fix: scope styles in MaAlert 

carlotamolintr:
Hi all,
This PR is to scope the styles from Ma-alert #401  because we are overlapping other component styles (as the RadioQuestions commented on the [issue before](https://github.com/holaluz/margarita/issues/409)). Also, I removed the tag defined previously to fix the MaOption because is not doing anything it should be solved with this fix.

### DIFF
--- a/src/components/MaAlert/MaAlert.vue
+++ b/src/components/MaAlert/MaAlert.vue
@@ -57,4 +57,4 @@ export default {
 }
 </script>
 
-<style src="./MaAlert.css"></style>
+<style scoped src="./MaAlert.css"></style>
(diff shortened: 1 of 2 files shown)


Samitier (approved):
nice! Is it a breaking change? Does anybody have `.icon` or `.alert` classes in their projects?

afontcu:
I'd define "using internal classes" it as relying on internals of the library, which can change change without further notice 😄  So I'd suggest issuing a patch or minor release, not a breaking change

edufarre (approved):

carlotamolintr:
Should I wait or I can merge this?

sergigraciaj:
U can merge it! : )

afontcu:
> Should I wait or I can merge this?

Feel free to merge it! Remember to give it a proper commit merge name. "Using internal classes" do not explain what are we fixing here! ;)

PR #415, opened 2021-05-06
feat: Add 'section' option to prop tag in Ma-Stack

sergigraciaj:


afontcu (approved):
If you add the `feat: ` block when writing the PR title there's less chances of forgetting it afterward ;)

sergigraciaj:
Soooooo true! xd

PR #420, opened 2021-05-18
feat: Add align prop ma heading

### LINKED ISSUE #403: Ma Heading should have align prop too (as Ma Text)?
We added align prop on Ma-Text but this wasn't added to MaHeading. Do you thing that this will be useful?

carlotamolintr:
Closes #403


src/components/MaHeading/MaHeading.stories.js
```diff
@@ -29,8 +35,8 @@ export default {
 const Template = (args, { argTypes }) => ({
   props: Object.keys(argTypes),
   template: `<div style="display: flex;">
-      <ma-heading v-bind="$props">I'M a heading</ma-heading>
-      <ma-heading v-bind="$props" style="margin-left:4rem; background-color:var(--color-blue-light)">
+      <ma-heading v-bind="$props" style=" width: 30rem" >I'M a heading</ma-heading>
```
edufarre, line 38:
```suggestion
      <ma-heading v-bind="$props" style="width: 30rem">I'M a heading</ma-heading>
```

edufarre (approved):

edufarre:
I believe we should add the `Close #403` in order to close the issue too

src/components/MaHeading/MaHeading.stories.js
Samitier: what if we move both these 30rem to the parent div to not repeat ourselves?

Samitier (approved):

src/components/MaHeading/MaHeading.stories.js
carlotamolintr: @Samitier, because to playing with controls you can see easier like this (as you have the blue background).

edufarre: Well... I did the first approach of `ma-text` and `ma-heading` and I was aiming to highlight the `negative margin crop` of the element with the `background-color: blue`, dunno if that's required/needed/desired anymore. 
Feel free to change it if you'd like.

carlotamolintr: Hi @edufarre, I mean that the align prop is easier to see with the background.  On my side, I don't need to remove it.

afontcu (approved):

PR #422, opened 2021-05-25
feat: Rename props color to tone

sergigraciaj:
To mantain consistency through all our components we renamed _color_ props to _tone_.
This implies a breaking change in our projects, but we think we can afford it in order to make components props consistent and easy to remember.

- [x] MaCard -> color to tone
- [x] MaPill -> color to tone
- [x] MaIcon -> iconColor to tone
- [x] MaAlert -> prop type to tone, value info to blue, value error to red, value success to green, value warning to yellow


alexcastelo95 (approved):
LGTM 🔥

edufarre:
If this is a breaking change, the commit message should be `BREAKING CHANGE:`, right?
It would be nice if before the squash and merge, we do write down which components have been changed, to then look back to this and clearly see which components' properties we do have to change or modify.

afontcu:
> It would be nice if before the squash and merge, we do write down which components have been changed

That would be great, so then it is easier to write a proper changelog notes: https://github.com/holaluz/margarita/releases/tag/v5.0.0 (this can be done when merging, or even after)

afontcu:
Should we also replace MaAlert's type with tone?

sergigraciaj:
I'm not sure about that, _type_ in _MaAlert_ is not just a plain color, is also the border and the icon. _Type_ has a semantic meaning right now, and that implies a mental mapping to know the info value is the blue one, so if we change the prop obviously we have to change the values too.
Then to mantain consistency in the use of the props we should review _category_ in _MaButton_ and I'm not sure if _type_ in _MaList_.
Other than that, would make sense to change _headerType_ to _headerTone_ in _MaModal_.

afontcu:
> I'm not sure about that, _type_ in _MaAlert_ is not just a plain color, is also the border and the icon

The prop is not called `color` but `tone`, right? :) The thing is that, as you pointed out, we're using semantic values for Alert, and visual values for the other components. Using a single criteria might benefit us along the road.

I'd say list's type and Button's category are not a different component's "tone", but something else. Same with modal header type.

afontcu:
Looks like some stories are missing some changes (seen on Chromatic visual tests that are not passing)

alexcastelo95 (approved):

carlotamolintr:
LGTM!!  just an small thing... Could be possible to add in the controls the selector as works now? It will be easier to know which are the options available. If it needs to be deployed asap may we can do an small task/issue..

PR #425, opened 2021-06-03
feat: Add prop space to MaList

sergigraciaj:
and modify css so the icon is not centered vertically


cillascillas (approved):
👌  Very useful, I'll use it right away! Thanks!

afontcu (approved):

PR #427, opened 2021-07-01
feat: Add MaStepper

sergigraciaj:
https://www.notion.so/holaluz/MaStepper-Q4-0dc358112f254abe8380a568866db04f#9face00ded5c4ddda1739f9a372a4522
https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=5587%3A208

image

Pretty much the same implementation as in https://github.com/holaluz/bonasera/blob/master/src/ui/components/FormWrapper/FormWrapper.vue

This component is not responsible of:
- The width of itself, as we got components that handles width like MaContainer or MaLayout
- The content inside every step

This component is responsible of:
- Adding a number for every step, and a line between them
- The separation between every step


src/components/MaStepper/MaStepper.vue
```diff
@@ -0,0 +1,32 @@
...
+  name: 'MaStepper',
+
+  props: {
+    /**
+     * Sets the size of the number
+     */
+    size: {
+      type: String,
+      default: 'big',
+      validator: (val) => ['big', 'small'].includes(val),
```
afontcu, line 18: If I'm not mistaken we usually use "large" instead of "big"

sergigraciaj: Makes sense, thx!

afontcu:
❓  not sure if bug or feature:


https://user-images.githubusercontent.com/9197791/124156719-41dfc400-da98-11eb-9855-5b0dadbbc388.mov

sergigraciaj:
Should we add a margin-left? The width of the stepper will include the numbers and content, and not just the content then.

src/components/MaStepper/MaStepper.vue
```diff
@@ -0,0 +1,30 @@
...
+  name: 'MaStepper',
+
+  props: {
+    /**
+     * Sets the size of the number
+     */
+    size: {
+      type: String,
+      default: 'large',
+      validator: (val) => ['xsmall', 'small', 'medium', 'large'].includes(val),
```
afontcu, line 18: Not entirely sure this matches the design spec for the stepper?

image

sergigraciaj: Modified the Figma removing not used variants and adding proper sizes. Added prop Tone to the component 👍

afontcu:
Given the figma spec, it looks like gap between number and content changes depending on the `size` prop, and I'm not entirely sure we're handling this in the implementation?

Also (following the same spec) – Line between numbers should also be white when `tone="white"`

sergigraciaj:
Fixed the line color.
I thought designers & frontends agreed that the size prop is just for the number, the content is defined by the consumer.

afontcu:
>  I thought designers & frontends agreed that the size prop is just for the number, the content is defined by the consumer.

I mean the gap between each number and its associated content (which you are already defining in the component)

icurdu:
The gap between the stepper and the content will be always the same (1rem) in the first iteration. We will see if we need to manage the space on the right for every breakpoint/variant on the future.

IMHO using ma-heading components for the numbers will be enough for managing the responsiveness.
[See what I mean](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=9181%3A8039)

cillascillas (approved):

cillascillas:
Thanks a lot to @icurdu and #claraibars for the design vision and support! Looking forward to test it in Bonasera @sergigraciaj! 😄

PR #430, opened 2021-07-08
feat: Update shadow tokens

afontcu:
spec: https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=5314%3A807

### DIFF
--- a/src/css/tokens.css
+++ b/src/css/tokens.css
@@ -1,6 +1,6 @@
 /**
  * Do not edit directly
- * Generated on Mon, 15 Feb 2021 14:56:40 GMT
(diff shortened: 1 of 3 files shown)


afontcu:
Note that, while this is technically a breaking change, we're not using shadow tokens anywhere yet, so we can skip the pain of releasing a major version.

alexcastelo95 (approved):
I think I used the shadow styles in a custom class in a PR the other day because the design was with the new tokens already 🤦🏽 

So, let's go for it! 🔥

sergigraciaj (approved):

PR #432, opened 2021-07-19
feat: Include gradient tone in Ma-text and Ma-heading component

carlotamolintr:
To do:

- [ ] : Make gradient tone also for MaHeading

### DIFF
--- a/src/components/MaHeading/MaHeading.css
+++ b/src/components/MaHeading/MaHeading.css
@@ -17,3 +17,10 @@
     margin-top: var(--bottom-crop);
   }
 }
+
+.gradient {
+  background: linear-gradient(90deg, var(--gradient-holaluz));
+  -webkit-background-clip: text;
+  -webkit-text-fill-color: transparent;
+  padding: 0;
+}
--- a/src/components/MaHeading/MaHeading.vue
+++ b/src/components/MaHeading/MaHeading.vue
@@ -54,11 +54,18 @@ export default {
     const tag = props.level === 'none' ? 'div' : `h${props.level}`
     const size = parent.$layout.getResponsivePropValue(props.size)
     const sizeStyles = headingSize[parent.$layout.currentBreakpoint][size]
+    const gradient = props.tone === 'gradient'
+
     const componentData = {
       staticClass: 'ma-heading',
+      class: {
+        gradient: gradient,
+      },
       style: {
         fontSize: sizeStyles['font-size'],
-        lineHeight: sizeStyles['line-height'],
+        lineHeight: gradient
+          ? 1.5 * sizeStyles['line-height']
+          : sizeStyles['line-height'],
         '--top-crop': sizeStyles['top-crop'],
         '--bottom-crop': sizeStyles['bottom-crop'],
         color: tones[props.tone],
--- a/src/components/MaHeading/MaHeading.vue
+++ b/src/components/MaHeading/MaHeading.vue
@@ -33,7 +33,7 @@ export default {
 
     /**
      * Sets the text element color tone
-     * @values white, red, pink, blue, green, yellow, gray-darker, gray-dark, gray
+     * @values white, red, pink, blue, green, yellow, gray-darker, gray-dark, gray, gradient
      */
     tone: {
       type: String,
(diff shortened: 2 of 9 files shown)


src/components/MaText/MaText.vue
```diff
@@ -81,6 +82,13 @@ export default {
         '--bottom-crop': sizeStyles['bottom-crop'],
         textAlign: props.align,
         color: tones[props.tone],
+        background: gradient
+          ? `linear-gradient(90deg, #e6007d, #f06c17 85%, #ffba03)`
+          : '',
```
carlotamolintr, line 87: I want to add the linear gradient when there is gradient BUT NOT with a ternary operator to avoid write empty or init. How can I do that? Cause I tried and it seems doesn't work...

carlotamolintr:
> I would add a new boolean prop gradient and do something like `color: props.gradient ? '' : tones[props.tone] ` so all the gradient stuff can be removed from color and tone files.
> And the class could be `ma-text--gradient: props.gradient` as we do for italic and bold, so we don't need to use a const.

Hi! Thanks for your comment. :) 
Regarding this, since I know, we want to use the prop tone to manage the gradient too.

src/components/MaHeading/MaHeading.vue
```diff
@@ -54,11 +54,18 @@ export default {
...
     const componentData = {
       staticClass: 'ma-heading',
+      class: {
+        gradient: gradient,
+      },
       style: {
         fontSize: sizeStyles['font-size'],
-        lineHeight: sizeStyles['line-height'],
+        lineHeight: gradient
+          ? 1.5 * sizeStyles['line-height']
```
afontcu, line 67: ⚠️  Do we really need to increase line-height depending on a tone-related prop?

afontcu:
We should also update `@values` on tone prop for both MaText and MaHeading: https://github.com/holaluz/margarita/blob/master/src/components/MaHeading/MaHeading.vue#L34-L37

src/components/MaHeading/MaHeading.vue
carlotamolintr: That's the solution that I found...😅 . But maybe there's another! I'll explain the issue... the gradient is cutting the text so without increassing line-height or padding the text looks like that: 
image
(cause of the background-clip property)

afontcu: oh, I see. However, then Headings/Texts with gradient will behave differently than regular Headings/Texts without gradient. Setting a higher line-height will neglect the negative margin we're setting in our texts, so we might have a hard time trying to create an aligned layout.  

But if everyone's on board, I'm on board, too 👍

src/tokens/color.js
```diff
@@ -45,4 +45,5 @@ module.exports = {
   white: {
     base: '#ffffff',
   },
+  gradientHolaluz: {},
```
afontcu, line 48: Is this actually needed/working?

src/components/MaHeading/MaHeading.css
```diff
@@ -17,3 +17,10 @@
     margin-top: var(--bottom-crop);
   }
 }
+
+.gradient {
+  background: linear-gradient(90deg, var(--gradient-holaluz));
+  -webkit-background-clip: text;
+  -webkit-text-fill-color: transparent;
+  padding: 0;
```
afontcu, line 25: Is it possible that `padding: 0` is unnecessary, here?

src/components/MaHeading/MaHeading.vue
carlotamolintr: Oh no, I did it alone... So, please if let's discuss other ways!

src/tokens/color.js
carlotamolintr: This should be gradient instead of gradientHolaluz. I just added here to allow the validation with the tone prop... But I guess I can remove it🤔

src/components/MaHeading/MaHeading.vue
alexcastelo95: I am not sure how to handle this. I understand it means that all our layouts with a MaHeading with gradient are going to look slightly different than the others...

If that's fine with all of us, it's fine with me.

carlotamolintr:
At least we decided not include the gradient tone on ma-text and ma-heading on our DS. But we agree to keep the two gradients that we currently have on our websites (rate page in dry and sign-in in maitai).

PR #433, opened 2021-07-19
fix: fix modal shadow overlay

alexcastelo95:
### Summary
In this PR I have replaced the shadow variables that were left out with this [PR](https://github.com/holaluz/margarita/commit/afb5625c8833a58b52f7f7670642661693b02149).

I've noted some of the ones I've changed to see if the design crew has anything to say about it. 🚀


afontcu (approved):
I guess it's fine as a quick fix, but we should def check if our current shadow usages (Card, Modal…) need to match our shadow design tokens (spoiler: I think they should 😂). It is weird to have shadow tokens and not even use them in Margarita hahaha

edufarre (commented):
`rgba...` 🤔 

Not sure if this is the best approach, we could try find a proper shadow level for the ones that use to be/reference the old shadow names 😄

alexcastelo95:
> I guess it's fine as a quick fix, but we should def check if our current shadow usages (Card, Modal…) need to match our shadow design tokens (spoiler: I think they should 😂). It is weird to have shadow tokens and not even use them in Margarita hahaha

I fully agree. I have the list with the components that use shadow to pass it through the design channel and let them tell us what to do with them.

edufarre (approved):

Issue #405, opened 2021-04-29
We need a way to do a layout where columns auto-fit


Samitier:
## What's the problem?

We have a missing feature in MaLayout that we desperately need. It's basically the "`display: flex` experience" 😂 

We need a way to set up columns that automatically fit their own width inside a row, like:

![image](https://user-images.githubusercontent.com/4160121/116520647-48549280-a8d3-11eb-9412-52dcb90dd9a4.png)

Right now, the closest thing we have in MaLayout is grid-auto-flow (the `*` operator), which always fills the whole width of the row, no matter the content of the columns.

![image](https://user-images.githubusercontent.com/4160121/116522377-6cb16e80-a8d5-11eb-87fc-d10561623498.png)

I would like to not change the API of MaLayout or having to create a different component for this, and I think we can get something working if we play with the `justify` prop.

## What do we do to fix it?

What if the `grid-auto-flow` operator works differently, depending on the `justify` prop? 

So basically this:

![image](https://user-images.githubusercontent.com/4160121/116523045-17299180-a8d6-11eb-8d72-3ffa10890648.png)

What do you think? It will be a breaking change for sure 😞

Samitier:
hmm if we use `justify: center` we will lose the opportunity to center the row, so maybe it's not a good idea. Maybe we need a new value for justify

edufarre:
What about allowing the `columns` property to have a value like `max-content`, thus keeping the same way approach as using `css grid` and not mixing different properties behaviour.

afontcu:
What about `<ma-layout>` without `columns` prop? Too magical? 😂

sergigraciaj:
> hmm if we use justify: center we will lose the opportunity to center the row, so maybe it's not a good idea. Maybe we need a new value for justify

Totally, `justify: expand`, ` justify:fit`? Or just boolean fit prop?

Samitier:
> What about allowing the `columns` property to have a value like `max-content`, thus keeping the same way approach as using `css grid` and not mixing different properties behaviour.

I like how it sounds, but it adds even more complexity for an already complex prop `columns` IMHO. I didn't want to touch that because I think it's on the verge to become unreadable 😂 .


> What about `<ma-layout>` without `columns` prop? Too magical? 😂 It'd be similar to Stack (it has no "columns/row" props, but it lays content vertically).

Then we wont be able to build it with other rows with a single `ma-layout`. For example, if we want a 12 col row before that one, we will need to do:
```html
<ma-stack>
[...]

PR #434, opened 2021-07-19
feat: ma-layout asterisk operator makes same-size columns

Samitier:
This one is a bit difficult to understand, so I will try my best:

## What's the problem?

The whole `ma-layout` component revolved around playing with `grid` column props. More explicitly, `grid-template-columns`, which [dictates the column widths of the child elements of the grid](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns).

However, there is a weird ""feature"" which makes this css prop unable to work properly when there is a `gap` in the grid and the sizes [are set in percentages](https://stackoverflow.com/a/45092180). Because of that, we ended up having to subtract the size of the gap with a `calc`, which probably hurted performance.

Moreover, the autoflow operator `*` didn't work properly when columns had different content. This detail was hidden in the examples because all children had the same content and, therefore, the same width.

## So what does this do exactly?

While trying to make all children of a `columns="*"` the same size I discovered this css prop: [`grid-auto-columns`](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-auto-columns), which lets us set a width for the columns when the grid has `grid-auto-flow: columns`.

Then I wondered why wasn't it all a `grid-auto-flow: columns` if we were basically only piling up rows with a `ma-stack`.

And THEN I discovered that if a column is less than `1fr` in size, the column will not fill all the width of the grid, just the proportional part. So all those operations with `calc` are redundant because we can do everything with `fr`. 

So I simplified everything to use only `grid-auto-flow: columns` and `grid-auto-columns` and `fr` units and seems like everything renders properly.

## Is this a breaking change?
Sadly, I think it is. And a difficult to fix one. The behavior of `columns="*"` its not the same anymore and columns will have all the same size now. Also, we completely change the way we render columns so it might happen that there are unexpected results. I tried in `dry-martini` and everything rendered properly though.

Fixes https://github.com/holaluz/margarita/issues/405, ~and maybe improves performance of ma-layout.~ nope, it's a [mozilla bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1591366).


src/components/MaColumns/MaColumns.spec.js
```diff
@@ -61,16 +61,19 @@ describe('MaColumns', () => {
...
   })
 
-  test('adds auto flow if columns has autoFlow operator', () => {
+  test('adds constant column width if columns has autoFlow operator', () => {
     const { contentWrapper } = renderComponent({ columns: '*' })
-    expect(contentWrapper).toHaveClass('has-auto-flow')
+    expect(contentWrapper).toHaveAttribute(
+      'style',
+      expect.stringContaining('1fr')
+    )
```
afontcu, line 69: just to get a bit fancy, there's `toHaveStyle`:


```suggestion
    expect(contentWrapper).toHaveStyle({ gridAutoColumns: '1fr' })
```

src/components/MaLayout/MaLayout.vue
```diff
@@ -118,29 +120,27 @@ function renderStack({ createElement, dom, gap, data }) {
...
-
-  grid.forEach((row, i) => {
-    const isLast = i === grid.length - 1
-    const children = dom.filter((c) => c.tag)
-    const rowChildren = children.slice(
-      lastIdx,
-      isLast ? children.length : lastIdx + row.length
+  const elements = dom.filter((c) => c.tag)
+  let i = 0
+  while (elements.length) {
```
afontcu, line 126: Could we drop a small comment here, to make clear what we're tying to achieve? Looks a bit convoluted and maybe this would help 👍

afontcu:
Thank you for the thorough explanation! Do you think this is worth releasing as a beta, to check against edge cases and also explore potential performance gains?

Samitier:
> Thank you for the thorough explanation! Do you think this is worth releasing as a beta, to check against edge cases and also explore potential performance gains?

You took the words from my mouth. I am not 100% confident that this will render exactly the same as we have in prod right now, so a beta would be perfect to check it. I'm going to do it and I'll check in dry/maitai/bonasera.

src/components/MaLayout/MaLayout.vue
Samitier: I tried 😂  does it make sense?

src/components/MaColumns/MaColumns.spec.js
Samitier: nicee 🔝

src/components/MaLayout/MaLayout.vue
afontcu: it does ❤️

Samitier:
everything works as expected in maitai, dry-martini and bonasera 🎉

afontcu (approved):

carlotamolintr (approved):

PR #436, opened 2021-07-23
docs: add ma-hidden documentation

carlotamolintr:
Hello all, 

I just did the ma-hidden documentation. 

Do we need to add examples too?

I wait for your comments


docs/components/layout/MaHidden.docs.mdx
```diff
@@ -1,3 +1,20 @@
...
+- **below**: Hides the content below the desired breakpoint. Whose values can be "desktop", "tablet" and "null.
+- **above**: Hides the content above the desired breakpoint. Whose values can be "mobile" , "tablet" and "null".
+
+There is another property that sets if content wrapper should be an inline element:
+
+- **inline:** This is a boolean property. Set as true, the html tag will be a `<div>` Set as false will be as `<span>`.
+
+## When to use it
+
+In that cases that we need to show or hide an element in a specific breakpoints
```
edufarre, line 20:
```suggestion
In those cases that we need to show or hide an element in specific breakpoints
```

edufarre (approved):
I believe the component is quite explanatory by itself, I don't think having examples would bring more value in this case 

🚀

alexcastelo95 (approved):
Looks good. Is it possible that we need to add an I-Frame to take us to Figma as well?

If not it can be done in another iteration.

alexcastelo95:
> Looks good. Is it possible that we need to add an I-Frame to take us to Figma as well?
> 
> If not it can be done in another iteration.

Come to think of it, maybe the MaHidden is difficult to explain in Figma. 😂

Issue #431, opened 2021-07-08
Nuxt-links with Ma-Buttons inside


sergigraciaj:
In some places in our repos we have Nuxt-Link with a Ma-Button inside, this is due that long ago we accepted "span" and "div" as a prop for tag property, then we modified the component to use "a" and "button" as tags.
But Nuxt-Link already generates an "a" tag, so a Ma-Button inside -"a" or "button" tags-  [is not valid HTML5](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element).

How do we solve this?
- Not using Nuxt-Link and just using a Ma-button with "a" tag? Using Nuxt-Link give us a boost on our performance by prefetching the links, so we lose this for internal links.
- Modify Ma-Button to use div as tag prop and add the Nuxt-Link as slot, somehow like a wrapper.
- Use classes from Margarita. "Classes" as we have different kind of buttons, if we do so any visual change in the Ma-Button should be applied twice.

Any other idea?

afontcu:
Would this fix the issue with no effort from our side?

https://router.vuejs.org/api/#v-slot-api-3-1-0

```vue
<nuxt-link
  v-slot="{ href, navigate, isActive }"
  to="/my-amazing-page"
  custom
>
  <!-- drop here any component to render it as a link -->
</nuxt-link>
```

what do you think about that? Using a slot we would remove the need of handling nuxt-links whatsoever. Simply wrap a link (or a button!) and we're good. We can use this already XD

sergigraciaj:
I'll try that, thanks.

afontcu:
did it work? @sergigraciaj

sergigraciaj:
I close this issue as we have a solution for this. 👋

PR #441, opened 2021-08-11
feat: Add MaLink

carlotamolintr:
This PR is to create the ma-link with the features listed below.

Features for the ma-link:

1. `category` prop with same values as ma-button without `rounded`. Notice that `no-background` makes the link inline (that is, "normal"). So this should be aligned with text. If we want to align the ma-link with a button vertically, we will use the ma-layout component with verticalALign = center.
2. Required `href` prop.
3. We can use the nuxt-link tag containing the ma-link when is necessary. See https://github.com/holaluz/margarita/issues/431#issuecomment-898285536

OUT OF SCOPE:
- Handling links inside of Alert: https://github.com/holaluz/margarita/blob/master/src/components/MaAlert/MaAlert.css#L48-L76
- Updating ma-button:
    1. make sure the `no-background` category for ma-button makes the button inline (that is, behave like a link). To confirm this is a breaking change.
    2. remove `tag` prop.


sergigraciaj (approved):

src/components/MaLink/MaLink.css
```diff
@@ -0,0 +1,80 @@
+.ma-link {
```
Samitier, line 1: do we really need this file? Can't we just use `MaButton.css`?

`<style scoped src="../MaButton/MaButton.css"></style>`

afontcu: There are differences between the two (link has no `disabled` state, for instance, a link is inline by default whereas a button is a block element, there is no "rounder" category…), so we can't simply import them, it would require some tweaks.

With that in mind, it is true that we could unify the "category-level styles" and share them between Links and Buttons, but felt like overextending for this PR (as we wanted to keep it simple and focus on the Link, we even [created another PR to add an additional category to Button](https://github.com/holaluz/margarita/pull/442)) – this would simply be a refactor, so we can always tackle that on a future PR without any API change.

src/components/MaLink/MaLink.vue
```diff
@@ -0,0 +1,38 @@
...
+    },
+  },
+
+  computed: {
+    linkStyles() {
+      return {
+        [`ma-link--${this.category}`]: this.category,
+      }
+    },
+  },
```
Samitier, line 34: why not interpolating directly in the template?

```html
<a :class="`ma-link--${ category } ma-link`" :href="href"><slot /></a>
```

Or at least making this computed a bit more clear

```suggestion
  computed: {
    cssClass() {
      return `ma-link--${this.category}`
    },
  },
```

src/components/MaLink/MaLink.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <a :class="[linkStyles, 'ma-link']" :href="href"><slot /></a>
+</template>
+
+<script>
+export default {
+  name: 'MaLink',
+
+  props: {
```
Samitier, line 9: what about `target="__blank"` ?

src/components/MaLink/MaLink.vue
afontcu: ok, i'll go ahead and make the change, but since this is a stylistic change I'd rather have a linter rule to check against or leave it to the discretion of authors xD

src/components/MaLink/MaLink.vue
afontcu: You can always set `target="_blank"` by defining it as an attribute, there's no need to register it as a prop to make it work

src/components/MaLink/MaLink.vue
```diff
@@ -0,0 +1,30 @@
+<template>
+  <a :class="`ma-link--${category} ma-link`" :href="href"><slot /></a>
```
Samitier, line 2: what about adding `rel=”noreferrer noopener”` if href is an absolute url or does not contain `holaluz.com`?

src/components/MaLink/MaLink.css
Samitier: 👍  I thought both had exactly the same styles.

afontcu: We hoped that was the case, too 😂 turns out subtle things change.

However, I'll explore the idea of having a `category.css` or something that contains the specifics, shared styles of each category. Maybe we come up with ~80-90% of shared styles, and we can cope with the differences in split files (`button.css` and `link.css`) 👍 good point

src/components/MaLink/MaLink.vue
Samitier: yeah but its kinda cute `<ma-link href="foo" in-new-tab />`. 

All target="_blank" should have the `rel="noopener"` too for security.

Samitier (approved):

src/components/MaLink/MaLink.vue
afontcu: Ah, that's a good point. We'd need to add the attribute if (1) link is targeting a new page, and (2) it is an absolute link, right? I'll take a look.

Fun fact: we could simply set `rel="noreferrer"` because [it implies](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer) "noopener" if it wasn't due to… internet explorer 😄  https://caniuse.com/rel-noopener

afontcu: done https://github.com/holaluz/margarita/pull/441/commits/aefdfed071aef67231720cc319a337eeddc633a0

Issue #419, opened 2021-05-18
[RFC] MaForm 


Samitier:
Hello!

Lately, trying to make a simple form in maitai I found plenty of limitations in margarita for doing form validations. So here I go!

This is all inspired by the frustrations of https://github.com/holaluz/margarita/issues/417, so maybe it's better if you take a look there before reading this.

## What is this trying to solve?

Basically, we are all using the dependency vuelidate to manage form validations. But we are using it differently depending on the project and also repeating a lot of boilerplate. I think that we should have a more streamlined solution to this.

Also, the way we are adding form validations (at least in maitai) is full of side effects, which makes the code super fragile:

```js
// a validations.js file
export default {
  couponCode: {
    isAvailableCoupon() {

      // wtf is this.coupon? this is using the scope of the component even though its in a different file
      const isAvailableCoupon = this.coupon.isAvailable

      if (!isAvailableCoupon) {

       // Why are we sending gtm events in a validation?
        this.$gtm.pushEvent({
          eventAction: 'invitations out',
          eventLabel: this.coupon.code,
        })
      }

      return isAvailableCoupon
    },
```
And redundant listeners:
```vue
<ma-text-field
      id="mgmCouponCode"
      v-model.trim="couponCode"
      class="input"
      :label="$t('submit.referral.label')"
      :placeholder="$t('submit.referral.placeholder')"
      :has-error="hasError('couponCode')"
      :error-message="getErrorMessage('couponCode')"
      @blur="validate('couponCode')"
    />
<!-- we always have to listen the blur event. Also, why hasError and validate are different functions? 
```
```js
 // and always watch this $v.$invalid prop (that is hidden by a mixin btw)
  watch: {
    '$v.$invalid': {
      handler(isInvalid) {
        if (isInvalid && this.code !== this.coupon.code) {
          this.setCoupon(ReferralCode())
        }
      },
    },
  },
```
So IMHO, we have to add a lot of boilerplate and questionable code to validate an input. Also, we have to depend on vuelidate instead of just using margarita for that. Margarita has a pseudo error managing (`error-message`/`has-error`) but it is not accessible or even usable without said boilerplate (see https://github.com/holaluz/margarita/issues/417)

## Ok, stop the rant. What do you propose?

I think we can simplify all this a lot. For a form, I think that what we all need is the ability to give a validator function to all of our inputs and also a variable that tells us if the whole form has any problem or not in order to let the user continue.

I'm going to do it with an example:

Imagine this form:

[...]

Issue #417, opened 2021-05-07
MaTextField: do not show error if field is focused or pristine


Samitier:
IMHO MaTextField doesn't work properly, from an ux pov. There are 2 issues:

I think we should never show the error if a ma-text-field is selected, regardless of what the prop `has-error` says. Its pretty unconfortable/anti-UX to tell the user they have errors on the input while they are writing:

![image](https://user-images.githubusercontent.com/4160121/117412924-a574c780-af15-11eb-8ec1-66b7f041aff8.png)

Same when the input is empty and has never been interacted with (pristine):

![image](https://user-images.githubusercontent.com/4160121/117413629-7b6fd500-af16-11eb-986b-5ae69686d3fd.png)

So why doesn't MaTextField enforce that by itself? Why do we need vuelidate or checking the blur event for this? IMHO, we should be able to do something like this and expect it to be usable out of the box:

```vue
<ma-text-field
    v-model="email"
   :has-error="isEmailValid"
   :error-message="'This email is not correct'"
/>

// computed
isEmailValid() {
    // a validation here that only checks if the email is correct or not, 
    // not if the input is blurred/focused.
}
```

Would this break any of the current implementations? For example: is there any form where we expect to show errors even if the user hasn't touch any fields?

I'm not going to enter on how redundant I find having 2 props for managing errors: `hasError` and `errorMessage`. Let's leave this for another issue :trollface:

Samitier:
I removed the mentions of vuelidate from this issue, its better to talk about that dependency in #419. This is only for improving the ux of `ma-text-field`.

afontcu:
>  I also wonder how necessary is Vuelidate to us, could we check for errors by our own? Should this happens in the use cases?
> I think that we should give feedback on every error an input could have like for example, required, minimum length, format...

We might have some default, simple validators (we actually do – we're using `input` after all! `<input type="email" />` is already a validation, and so is `<input type="text" pattern="[A-Za-z]{5}" />`), but how about validating for a valid power value? Or a valid consumption? A valid CUPS? Postal Code? Do we need to bake all this knowledge inside of MaTextField?

Maybe we could have our own set of validators, that's true. However I'm not sure Margarita is the right place for them to leave. Kali?

Samitier:
closed by https://github.com/holaluz/kalimotxo/pull/86

PR #443, opened 2021-08-16
[PoC] MaForm

Samitier:
Hello! This is just a proof of concept to see if my idea was feasible or not. The code is super rushed and probably could be way better written, but let's focus on the main idea to see if you like it or not.

## What is this?

Basically is a way to streamline form validation and form UX so that all of our forms work the same way. I tried to do the simplest, most general solution I could think of in order to see if all of our use cases would fit in it. That's why I ended up not adding any dependency, since sometimes we use vuelidate, sometimes veevalidate, sometimes nothing. And honestly, after checking it for a while I don't see a real necessity to use one or another (let's debate it tho!).

## What does this do?

This tries to solve some problems I found we have with our forms:
 - The forms do not submit when pressing enter.
 - We normally disable the submit button when there is an incorrect field. This is bad UX, we should let the user press it and then show the incorrect field.
 - The form validations are handled differently in every project. There is no way of sharing validators between projects. Some of the form validations contain side effects that are hard to maintain.
 - There is a lot of boilerplate in order to create a working form. We have to manually do basic UX stuff that should be done by margarita directly (for example, not showing error messages if an input is focused or pristine).

And so far it solves all of them. Maybe there are stuff other projects need that are not reflected in here, so let's talk it.

## How do we do it

This introduces three concepts:
 -  **MaForm**: a root form element that handles the validation of the whole form. Aka "a group of `MaFormFields`".
 -  **MaFormField**: handles validations of an input element. Basically, a wrapper of `MaTextField` that validates on blur and shows/hides the validation message when the user is writing or when the input element has not been touched.
 - **Validator**: a function that receives a value and returns `''` if the value is correct, or an error key `errror-whatever` if the input is incorrect. That error key will be the one we use to show error messages in i18n.

### Example

This proof of concept renders this form:

image

The basic markup/API is this:

```vue
<template>
<ma-form @submit="onSubmit">
    <ma-stack space="medium">
      <ma-form-field
        label="Full name"
        v-model="name"
        :validators="[validators.isRequired]"
      />
      <ma-form-field
        label="Email address"
        v-model="email"
        :validators="[validators.isRequired, validators.isEmail]"
      />
      <ma-form-field
        label="Phone number"
        v-model="phone"
        :validators="[validators.isRequired, validators.isPhoneNumber]"
      />
      <ma-button type="submit">Submit</ma-button>
    </ma-stack>
  </ma-form>
</template>
<script>
import * as validators from '@margarita/validators'

export default {
 computed: {
    validators() {
      return validators
    },
  },
  data() {
    return {
      name: '',
      email: '',
      phone: '',
    }
  },
  methods: {
    onSubmit() {
      // sending stuff to core or wherever is needed. This will only execute when all form fields are valid.
    },
  },
}
</script>
```

Please, try the [deploy preview](https://601199d127273f0021b037b0-fmtmrbzalg.chromatic.com/?path=/story/components-form--form) because most of the stuff discussed here is hard to describe if you don't see it in person. If you have any question of how I did it, it's pretty stupid. Basically `MaForm` is searching for `MaFormFields` in its children and checking whether they are invalid or not. If the `MaForm` has any child that is invalid it won't emit a submit button, it will just scroll to the first input with an error.

## Caveats

There are some stuff that this can't do, even though I don't think it would be specially difficult to add:
 - Async validators. For example, validations that have to check things in core.
 - The way of importing validators is a bit meh because we can't use them directly in the template, we have to use a computed. Maybe `validators` could accept a string and let `ma-form-field` handle the import.
 - This only validates `ma-text-fields`. Adding other fields would require extending the `ma-form-field` wrapper. I think there is no problem with `ma-select`, for example. But things like `ma-option` start being a bit weird, since we don't have an "invalid state" for them.
 - Full-form validations. That is, validations that happen AFTER we press submit. In my mind makes sense to just do them in the component that renders the `ma-form` because what we do after we press submit might vary depending on the use case. A good example for this is the use case I used for this proof of concept: "saving a lead". The lead email can be non-deliverable, but we will only know that if core returns a 400. So for example, we can show a message under the `ma-text-field` of the email if we submit an incorrect email without having to implement extra stuff:

```vue
<template>
<ma-form @submit="saveLead">
    <!-- we add a custom validator "isEmailDeliverable" -->
     <ma-form-field :validators"[isRequired, isEmail, isEmailDeliverableValidator]" ... />
</template>

<script>
...
  methods: {
     async saveLead() {
           try { await core.saveLead(...)}
           catch { this.isEmailDeliverable = false }
     },
     isEmailDeliverableValidator() {
         return this.isEmailDeliverable ? '' : 'error-email-non-deliverable' 
     }
  }
```

## Stuff I think would be nice to add after reading about form UX

- Animations to the validation messages. Maybe we improve the UX if the validation icon and text has some kind of movement.
- "Check" markers if fields are correct. Maybe marking the fields that are filled correctly would improve UX too.
- A "toast" message if we try to send an invalid form, not only rely on the input error messages to show errors.

What do you think? Can you think of any use case where this is completely unusable? Or some important feature this thing does not fix? Any comment would be amazing 😃

### DIFF
--- a/src/components/MaForm/MaForm.stories.js
+++ b/src/components/MaForm/MaForm.stories.js
@@ -1,0 +1,55 @@
+import MaForm from './MaForm'
+import * as validators from './_validators/validators'
+
+export default {
+  title: 'Components/Form',
+  component: MaForm,
+}
+
+const Template = (args, { argTypes }) => ({
+  props: Object.keys(argTypes),
+  computed: {
+    validators() {
+      return validators
+    },
+  },
+  data() {
+    return {
+      name: '',
+      email: '',
+      phone: '',
+      isSubmitted: false,
+    }
+  },
+  methods: {
+    onSubmit() {
+      this.isSubmitted = true
+      setTimeout(() => {
+        this.isSubmitted = false
+      }, 2000)
+    },
+  },
+  template: `<ma-form @submit="onSubmit" style="width: 300px">
+    <ma-stack space="medium">
+      <ma-form-field
+        label="Full name"
+        v-model="name"
+        :validators="[validators.isRequired]"
+      />
+      <ma-form-field
+        label="Email address"
+        v-model="email"
+        :validators="[validators.isRequired, validators.isEmail]"
+      />
+      <ma-form-field
+        label="Phone number"
+        v-model="phone"
+        :validators="[validators.isRequired, validators.isPhoneNumber]"
+      />
+      <ma-button type="submit">Submit</ma-button>
+      <ma-alert v-if="isSubmitted" tone="green">Form submitted successfully!</ma-alert>
+    </ma-stack>
+  </ma-form>`,
+})
+
+export const Form = Template.bind({})
--- a/src/components/MaForm/MaForm.vue
+++ b/src/components/MaForm/MaForm.vue
@@ -13,13 +13,6 @@ export default {
         (c) => c.$options._componentTag === maTextTag && c.hasError
       )
     },
-    submitElement() {
-      const maButtonTag = 'ma-button'
-      return this.$children.find(
-        (c) =>
-          c.$options._componentTag === maButtonTag && c.$attrs.type === 'submit'
-      )
-    },
   },
   methods: {
     onSubmit() {
--- a/src/components/MaForm/MaForm.vue
+++ b/src/components/MaForm/MaForm.vue
@@ -1,0 +1,39 @@
+<template>
+  <form ref="formElement" @submit.prevent="onSubmit">
+    <slot />
+  </form>
+</template>
+
+<script>
+export default {
+  computed: {
+    inputFieldsWithError() {
+      const maTextTag = 'ma-form-field'
+      return this.$children.filter(
+        (c) => c.$options._componentTag === maTextTag && c.hasError
+      )
+    },
+    submitElement() {
+      const maButtonTag = 'ma-button'
+      return this.$children.find(
+        (c) =>
+          c.$options._componentTag === maButtonTag && c.$attrs.type === 'submit'
+      )
+    },
+  },
+  methods: {
+    onSubmit() {
+      if (this.inputFieldsWithError.length) {
+        this.inputFieldsWithError.forEach((i) => {
+          i.hasBeenBlurred = true
+        })
+        this.inputFieldsWithError[0].$el.scrollIntoView({ behavior: 'smooth' })
+      } else {
+        this.$emit('submit')
+      }
+    },
+  },
+}
+</script>
+
+<style></style>
--- a/src/components/MaForm/_components/MaFormField.vue
+++ b/src/components/MaForm/_components/MaFormField.vue
@@ -1,0 +1,73 @@
+<template>
+  <ma-text-field
+    v-model="model"
+    :placeholder="placeholder ? placeholder : label"
+    :has-error="!!errorMessage"
+    :error-message="errorMessage"
+    :label="label"
+    @focus="onFocus"
+    @blur="onBlur"
+  />
+</template>
+
+<script>
+export default {
+  props: {
+    value: {
+      type: [String, Number],
+      required: true,
+    },
+    label: {
+      type: String,
+      required: true,
+    },
+    placeholder: {
+      type: String,
+      default: '',
+    },
+    validators: {
+      type: Array,
+      default: () => [],
+    },
+  },
+  data() {
+    return {
+      isFocused: false,
+      hasBeenBlurred: false,
+    }
+  },
+  computed: {
+    errorMessages() {
+      return this.validators
+        .map((validator) => validator(this.model))
+        .filter((v) => v)
+    },
+    hasError() {
+      return !!this.errorMessages.length
+    },
+    errorMessage() {
+      if (this.isFocused || !this.hasBeenBlurred || !this.hasError) return ''
+      const [errorMessage] = this.errorMessages
+      if (this.$t) return this.$t(errorMessage)
+      return errorMessage
+    },
+    model: {
+      get() {
+        return this.value
+      },
+      set(value) {
+        this.$emit('input', value)
+      },
+    },
+  },
+  methods: {
+    onBlur() {
+      this.hasBeenBlurred = true
+      this.isFocused = false
+    },
+    onFocus() {
+      this.isFocused = true
+    },
+  },
+}
+</script>
--- a/src/components/MaForm/_components/MaFormField.vue
+++ b/src/components/MaForm/_components/MaFormField.vue
@@ -1,21 +1,27 @@
 <template>
-  <ma-text-field
+  <component
+    :is="component"
     v-model="model"
-    v-bind="$attrs"
+    v-bind="{ ...$attrs, ...$props }"
     :placeholder="placeholder ? placeholder : label"
     :has-error="!!errorMessage"
     :error-message="errorMessage"
     :label="label"
-    @focus="onFocus"
     @blur="onBlur"
-  />
+    >{{ label }}</component
+  >
 </template>
 
 <script>
 export default {
   props: {
+    type: {
+      type: String,
+      validator: (v) => ['text', 'radio', 'checkbox', 'select'].includes(v),
+      default: 'text',
+    },
     value: {
-      type: [String, Number],
+      type: [String, Number, Boolean],
       required: true,
     },
     label: {
@@ -33,11 +39,21 @@ export default {
   },
   data() {
     return {
-      isFocused: false,
       hasBeenBlurred: false,
     }
   },
   computed: {
+    component() {
+      switch (this.type) {
+        case 'radio':
+          return 'MaOption'
+        case 'checkbox':
+          return 'MaOption'
+        case 'select':
+          return 'MaSelect'
+      }
+      return 'MaTextField'
+    },
     errorMessages() {
       return this.validators
         .map((validator) => validator(this.model))
@@ -47,7 +63,7 @@ export default {
       return !!this.errorMessages.length
     },
     errorMessage() {
-      if (this.isFocused || !this.hasBeenBlurred || !this.hasError) return ''
+      if (!this.hasBeenBlurred || !this.hasError) return ''
       const [errorMessage] = this.errorMessages
       if (this.$t) return this.$t(errorMessage)
       return errorMessage
@@ -64,10 +80,6 @@ export default {
   methods: {
     onBlur() {
       this.hasBeenBlurred = true
-      this.isFocused = false
-    },
-    onFocus() {
-      this.isFocused = true
     },
   },
 }
--- a/src/components/MaForm/_components/MaFormField.vue
+++ b/src/components/MaForm/_components/MaFormField.vue
@@ -1,6 +1,7 @@
 <template>
   <ma-text-field
     v-model="model"
+    v-bind="$attrs"
     :placeholder="placeholder ? placeholder : label"
     :has-error="!!errorMessage"
     :error-message="errorMessage"
--- a/src/components/MaForm/_validators/validators.js
+++ b/src/components/MaForm/_validators/validators.js
@@ -1,0 +1,11 @@
+export const isRequired = (value) => (value ? '' : 'error-required')
+
+export const isEmail = (value) => {
+  const isEmailValidEx = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+  return isEmailValidEx.test(value.toLowerCase()) ? '' : 'error-email-invalid'
+}
+
+export const isPhoneNumber = (value) => {
+  const isPhoneValidEx = /^[6789]{1}[0-9]{8}$/g
+  return isPhoneValidEx.test(value) ? '' : 'error-phone-invalid'
+}
--- a/src/components/MaForm/examples/example-form.vue
+++ b/src/components/MaForm/examples/example-form.vue
@@ -1,17 +1,21 @@
 <template>
-  <ma-form class="form" @submit="onSubmit">
-    <ma-stack space="medium">
-      <slot />
-      <ma-button type="submit">Submit</ma-button>
-      <ma-alert v-if="isSubmitted" tone="green">
-        Form submitted successfully!
-      </ma-alert>
-    </ma-stack>
-  </ma-form>
+  <div class="container">
+    <ma-form class="form" @submit="onSubmit">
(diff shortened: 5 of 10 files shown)


afontcu:
Thanks for leading on this! ❤️ 

Some questions:

1. You mentioned it, but I'm not entirely sure how does this integrate with other form elements? For instance, would we end up with a `MaFormCheckbox` or similar? Does it also mean `MaTextField` would be deprecated, and only used internally to provide a UI element for the `MaFormField`? vee-validate et al "look for" elements with a `v-model` binding and then control them, not sure how easy or hard this would be to replicate.
2. What would a cross-validation scenario look like? Say, "confirm your password/email" or similar.

---

Also, some feedback on the desired UX:


> We normally disable the submit button when there is an incorrect field. This is bad UX, we should let the user press it and then show the incorrect field.

Is it bad UX? I always thought having a disabled button + **good feedback on invalid fields** was the way to go. Happy to switch to enabled + feedback on click, but I'd like to be 100% sure this is what expected forms to behave.

> validates on blur and shows/hides the validation message when the user is writing or when the input element has not been touched

Do we always want this behavior? Let's say I type "asd" on the email field. I blur, then I get the error message. Then I focus it again (the error message goes away), I type something else, and then I need to blur again in order to check if it right or wrong: I don't get the feedback immediately (that is, the minute I type a valid email).

Also: Let's say I have 2 validators. I type something that makes the first one pass. Then I blur, I type something else, which makes the first pass and the second fail. Then I blur and I see some other string appearing. With an always-visible error message I could see the diff when typing, not after blurring.

(_That being said, maybe we should agree on the desired UX along with other design folks, and come up with a wishlist_?)

---

My main gripe here is that we're kind of reinventing the wheel (to some extent!). Tools such as `vee-validate` provide almost everything you identified and must-haves and should-haves out of the box, and it would prevent us from creating/maintaining validations, integrating the Form/FormField components with every form component margarita features, and so on.

(_Please notice that these points do not invalidate the idea of a MaForm – I'm just trying to make sure we're not missing anything important/better before we invest on the idea_ 😉 Some of the points you mentioned such as boilerplate and heterogeneous solutions… are totally valid and worth addressing)

Samitier:
> 1. You mentioned it, but I'm not entirely sure how does this integrate with other form elements? For instance, would we end up with a `MaFormCheckbox` or similar? Does it also mean `MaTextField` would be deprecated, and only used internally to provide a UI element for the `MaFormField`? vee-validate et al "look for" elements with a `v-model` binding and then control them, not sure how easy or hard this would be to replicate.

The `MaFormField` wrapper only exists because the API of `MaTextField` is a bit 💩 . For example, it does not handle showing the error message only when the field is not focused or have never been touched. It's the same wrapper we use in `dry-martini`.

I don't think we even need a wrapper for `MaFormCheckbox` since it does not have error states. Maybe for `ma-select` we need a wrapper too. Or maybe we can use `ma-form-field`, since it will have exactly the same logic. I can try to add selects in this PoC and see how easy it is.

We can also make a massive breaking change and refactor MaTextField and MaSelect, but yeah... scary 😂 

Or you think we should allow to have input types that are not in margarita in our forms? How can we standarize their validations/styles/ux then?

> 2. What would a cross-validation scenario look like? Say, "confirm your password/email" or similar.

You can add any validation you want. For example, a "repeat password" scenario:

```vue
<template>
     [...]
      <ma-form-field
        label="Password"
        v-model="password"
        :validators="[validators.isPassword]"
      />
      <ma-form-field
        label="Repeat password"
        v-model="repeatPassword"
        :validators="[validators.isPassword, isEqual]"
      />
      [...]
</template>
<script>
import * as validators from '@margarita/validators'

export default {
 computed: {
    validators() {
      return validators
    },
    isEqual(value) {
       return value === this.password ? '' : 'error-password-mismatch'
    }
  },
 [...]
```
> Is it bad UX? I always thought having a disabled button + **good feedback on invalid fields** was the way to go. Happy to switch to enabled + feedback on click, but I'd like to be 100% sure this is what expected forms to behave.

The problem with that is that we don't show feedback for required fields if they are empty. We simply don't let the user continue with nothing showing in the input and with the button disabled. If you try the deploy preview you will see that if you don't touch the form you will get feedback for the required fields when pressing continue.

> Do we always want this behavior? Let's say I type "asd" on the email field. I blur, then I get the error message. Then I focus it again (the error message goes away), I type something else, and then I need to blur again in order to check if it right or wrong: I don't get the feedback immediately (that is, the minute I type a valid email).
> 
> Also: Let's say I have 2 validators. I type something that makes the first one pass. Then I blur, I type something else, which makes the first pass and the second fail. Then I blur and I see some other string appearing. With an always-visible error message I could see the diff when typing, not after blurring.

Completely valid opinion. To me all solutions are ok as long as we do that consistently in all of our forms. Not having some forms behaving one way and others behaving differently. I think the design people will be able to discuss this more objectively than us.

> (_That being said, maybe we should agree on the desired UX along with other design folks, and come up with a wishlist_?)

Exactly this 😃

> My main gripe here is that we're kind of reinventing the wheel (to some extent!). Tools such as `vee-validate` provide almost everything you identified and must-haves and should-haves out of the box, and it would prevent us from creating/maintaining validations, integrating the Form/FormField components with every form component margarita features, and so on.

Yes, I was kinda divided about that. As you can see the API of this poc is completely inspired by vee-validate `form` and `field`, so switching to it should not change the outcome that much. 

I didn't use a library because honestly this thing is so trivial that is a bit overkill to justify the cost of adding it 😂 . IMHO we don't need much more than this to validate stuff, but if there are features that we need from veevalidate or if you think the code would be more robust, let's switch no prob.

afontcu:
Sorry I missed yet another common use case I wanted to ask you about: what about input groups? Such as having six text inputs to introduce power values, and validating them against business rules (they must be ascending, over 15000… and so on). Also, having a group of three checkboxes where selecting one is required.



> Or you think we should allow to have input types that are not in margarita in our forms? How can we standarize their validations/styles/ux then?

I was thinking of that but nah you're right, we should not.

> I didn't use a library because honestly this thing is so trivial that is a bit overkill to justify the cost of adding it 😂 . IMHO we don't need much more than this to validate stuff

That's a completely valid point! I think we should actually assess what are the goodies of vee-validate, and see what porting them to a custom solution would mean in both terms of building + documenting but also maintaining. Having a external lib is a two-edged sword: you get maintenance+docs for free, but you get coupled to an external solution. Using an anti-corruption layer would help, though.

Note, for instance, how complex defining what required means is: https://vee-validate.logaretm.com/v3/guide/required-fields.html#required-fields

vee-validate offers some cool goodies (in form of slot-scoped variables) such as `pristine`, `invalid`, `pending`, and so on, that might be tough to implement properly. The `handleSubmit` method is quite cool, too, because it revalidates the whole form before submitting. This makes sure people don't cheat my tweaking stuff in the devtools. It also [handles a11y](https://vee-validate.logaretm.com/v3/guide/state.html#aria-attributes) for free.

Again, I'm not 100% sold on vee-validate, I just hope we find a really good case for creating our own wrappers 😂 (it kind of make sense because vee-validate must be abstract enough to everyone, but we do not).

Samitier:
> Sorry I missed yet another common use case I wanted to ask you about: what about input groups? Such as having six text inputs to introduce power values, and validating them against business rules (they must be ascending, over 15000… and so on).

Maybe something like this?

```js
  data() {
    return {
      powers: [0, 0, 0, 0, 0, 0],
    }
  },
  methods: {
    powerValidator(minValue) {
      return (v) => {
        if (v < 15000) return 'Value should be greater than 15000'
        if (v <= minValue) return `Value should be greater than ${minValue}`
        return ''
      }
    },
  },
  template: `
    <ma-form>
        <ma-form-field
          v-for="(power, i) in powers"
          :label="'Power ' + i"
          v-model="powers[i]"
          type="number"
          :validators="[powerValidator(i ? powers[i-1] : 0)]"
        />
    </ma-form>
`,
```
You can check it live [here](https://601199d127273f0021b037b0-hyrscymywi.chromatic.com/?path=/story/components-form-examples--form-group-example).

> Also, having a group of three checkboxes where selecting one is required.

We can't do that because we don't have an "errored" checkbox in the design system, no? But we should 100%.

> Again, I'm not 100% sold on vee-validate, I just hope we find a really good case for creating our own wrappers 😂 (it kind of make sense because vee-validate must be abstract enough to everyone, but we do not).

I think even if we use something like vee-validate we will have to wrap our components to hide implementation details from vee-validate. Maybe now we can decide an api we all like, an UX we all like and then decide if it's easier/cleaner to implement with vee-validate or not. We can keep coming up with examples. I'm specially worried of async validations, but I don't know any real use case that has one.

We can also try to implement something similar (a `form` and `form-field` component combo) using vee-validate to see if it gets better.

davidbretos:
Hi there! First PD to join the conversation here, although we've been talking about it on Slack with @Samitier and @icurdu separately.

I'm a bit worried that we are going for a solution that might be a bit of an overkill. I'm probably missing some important technical aspects that introduce some constrains, but this PoC has quite a number of things to change and probably we could break them down to separate initiatives rather than try to solve them all at once.

From what I can recap after reading the thread and talking to @Samitier:

- **Disabled states of buttons** depending on what's pristine, has been filled out, has errors... This is an interesting debate, but is it possible to separate it from how should we deal with error states and error feedback on form components? Or technically we have to address it at the same time as error states and feedback? 
- **Error feedback when blur/focus:** from a UX point of view, it would make sense that the first error message shows up as soon as the user blurs the field (if it's a `ma-text-field`), remains visible if the user focuses again and disappears when the value is valid, as the user is typing.
- **Wrapper for form validation**: this can be tricky and maybe a bit overkill. Could we have a validator for the different component types that need validation? For example, what happens with range, that doesn't have validation?
- **Error messages in groups of components related to each other**: the way we deal with error feedback right now means that we show an error  message for each single field that has an error. Let's say that a step of the form has four `ma-text-field` with values related to each other: right now we would show an error message for each field, with is the equivalent of screaming to the user. We can probably do a better job by grouping them into a single error message. We'd probably need the next step of the component - we have the atoms (`ma-text-field`) and maybe we need the the molecules (groups of `ma-text-field`, or whatever other form component).

As you can see, this is a lot of stuff and probably we can work on this step by step, rather than trying to solve it all at once. We can discuss this in the Design System meeting tomorrow.

Samitier:
Hi @davidbretos 👋 

> this PoC has quite a number of things to change and probably we could break them down to separate initiatives rather than try to solve them all at once.

Yes! In fact, they are broken down in two issues: https://github.com/holaluz/margarita/issues/419 and https://github.com/holaluz/margarita/issues/417

This PoC was only to start the debate with something visual that we can tweak to try things. We can focus on what you think its more important.

> * **Disabled states of buttons** depending on what's pristine, has been filled out, has errors... This is an interesting debate, but is it possible to separate it from how should we deal with error states and error feedback on form components? Or technically we have to address it at the same time as error states and feedback?

For me, makes sense to divide it in 2: field validations and form validations.

Field validations could be: what is the UX when a single field is incorrect in a form?
Form validations could be: what happens when a form has at least an invalid field and a user tries to submit it?

There is a small problem, though. There are field validations that should happen AFTER the form is sent and maybe dividing it like that we will forget those. For example: what happens when a field is required? Do we let the user press "send" and then show the error under the field? Should we not let the user continue and assume they will find the missing field? Or for example, errors that happen when we cannot save a form for an external reason (i.e. we can't save a lead because the email does not exist). As long as we touch those edge cases it's fine for me!

> * **Error feedback when blur/focus:** from a UX point of view, it would make sense that the first error message shows up as soon as the user blurs the field (if it's a `ma-text-field`), remains visible if the user focuses again and disappears when the value is valid, as the user is typing.

❤️ 

> * **Wrapper for form validation**: this can be tricky and maybe a bit overkill. Could we have a validator for the different component types that need validation? For example, what happens with range, that doesn't have validation?

We should decide which form fields need validation and which do not. As you say, I don't think range needs it. 

The wrapper thing is only to not change `ma-text-field` and `ma-select` on every project that uses margarita, but maybe we should fix those components instead of introducing new ones 🤔 

> * **Error messages in groups of components related to each other**: the way we deal with error feedback right now means that we show an error  message for each single field that has an error. Let's say that a step of the form has four `ma-text-field` with values related to each other: right now we would show an error message for each field, with is the equivalent of screaming to the user. We can probably do a better job by grouping them into a single error message. We'd probably need the next step of the component - we have the atoms (`ma-text-field`) and maybe we need the the molecules (groups of `ma-text-field`, or whatever other form component).

Yes! A `form-group` component would be interesting. It would be nice finding a real form we have in any of our apps that would benefit from that to have a practical example.

afontcu:
> Let's say that a step of the form has four ma-text-field with values related to each other: right now we would show an error message for each field, with is the equivalent of screaming to the user. We can probably do a better job by grouping them into a single error message. We'd probably need the next step of the component - we have the atoms (ma-text-field) and maybe we need the the molecules (groups of ma-text-field, or whatever other form component).

We are already handle this in some places by splitting "UI errors" (that is, "X field is required") from "Domain errors" (that is, "These power values are not valid because blah blah"). The way we're handling this, though, is [not nearly as good as it should](https://github.com/holaluz/lumo360/blob/master/frontend/src/ui/pages/power-change/components/power-change-form.vue#L118-L132), which is something that kind of triggered this discussion.

vee-validate kind of [solves this by creating an abstract wrapper](https://vee-validate.logaretm.com/v3/advanced/cross-field-validation.html#cross-field-validation) (ValidationObserver) that gives full flexibility when it comes to (1) the number and type of inputs, (2) the error placement, and (3) the error trigger

Samitier:
> We are already handle this in some places by splitting "UI errors" (that is, "X field is required") from "Domain errors" (that is, "These power values are not valid because blah blah"). The way we're handling this, though, is [not nearly as good as it should](https://github.com/holaluz/lumo360/blob/master/frontend/src/ui/pages/power-change/components/power-change-form.vue#L118-L132), which is something that kind of triggered this discussion.

How do you render these domain errors when multiple inputs are involved? With a `ma-alert`? Can we see an example of a form like this so that we can take it into account when designing the solution?

afontcu:
> How do you render these domain errors when multiple inputs are involved? With a `ma-alert`? Can we see an example of a form like this so that we can take it into account when designing the solution?

Yep, you have it right there in the link, a few lines above 😉  https://github.com/holaluz/lumo360/blob/master/frontend/src/ui/pages/power-change/components/power-change-form.vue#L38-L49

edufarre:
👋🏽 Heya!! Looks really promising @Samitier ! However, I do not how I feel about the way we pass the validations, did you try the approach suggested earlier on this PR using the `veeValidate` lib? The form wrapper is quite useful, the scoped slots are intuitive and they handle the different input states easily. 
There is one little thing I don't really like about this approach, and I know it is probably because of me, it is the `type` property for `form-field` component, in my opinion, I'd rather use `slots` and then type in the template the margarita input we would like to use, thus making super easy to see/understand which component we are actually using there.

Great job 💪🏽 🎉

Samitier:
Implemented here -> https://github.com/holaluz/kalimotxo/pull/86

PR #444, opened 2021-08-16
fix: Ma-List keeps additional data

cillascillas:
### Context
If we add an additional class to `Ma-List` it is not kept to the `ul`/`ol` element that is created.
In order to make easier customizing or using the element in combined layouts (_i.e. adding an extra class_), `data` is added to the parent created element.


sergigraciaj (approved):
Awesome!

afontcu (approved):

PR #445, opened 2021-08-18
feat(MaOption): Add 'hasError' prop

Samitier:
image

The states of MaOption were not the same as the ones in [Watt](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=5314%3A1399). There were problems with the normal and focused states of `radio` and `checkbox`.

Also, there was an entire state missing: the error (for card, option and checkbox).

I also felt the need to cleanup all that messy css because it was impossible to understand anything 😂 . Sorry if it makes the review more difficult.


src/components/MaOption/MaOption.css
```diff
@@ -1,108 +1,168 @@
...
-.ma-selector-card .input {
-  &:not(:disabled) {
-    & ~ .description {
-      transition: border 0.2s ease;
-    }
-    &:focus {
-      & ~ .description {
-        border-color: var(--color-pink-light);
-      }
+.ma-card {
```
afontcu, line 1: We already have a `MaCard` component, so this is potentially misleading. Maybe `.ma-optin--card`?

Samitier:
I added 3 pages in storybook with the different states of the component

afontcu:
> I added 3 pages in storybook with the different states of the component

For extra confidence, you might want to add them in master first (creating a `docs:` PR or something) so then you could get a diff here. Otherwise, if the visual testing offered by storybook is enough, then 👍

sergigraciaj (approved):

afontcu (approved):
LGTM!

IMHO the merge commit should say something like `feat(MaOption): Add 'hasError' prop`, which is the actual API change for this PR :)

Samitier:
> For extra confidence, you might want to add them in master first (creating a `docs:` PR or something) so then you could get a diff here. Otherwise, if the visual testing offered by storybook is enough, then 👍

https://github.com/holaluz/margarita/pull/447

> IMHO the merge commit should say something like feat(MaOption): Add 'hasError' prop, which is the actual API change for this PR :)

Done

PR #446, opened 2021-08-19
feat(MaButton): Warn when using tag prop on a MaButton

### LINKED ISSUE #281: Make ma-button an anchor if paramerer href exists

afontcu:
Closes #281

This will help us identify and take action on usages of a deprecated prop.

## Potential replacements of `<ma-button tag="a">`

```vue
<ma-link href="#">a link that looks like a button</ma-link> <!-- (same as category="primary") -->

<ma-link href="#" category="no-background">a link that looks like a link</ma-link>
```


Samitier (approved):
🔝

PR #451, opened 2021-09-09
feat(MaStepper): Add space prop

sergigraciaj:
Add space prop in order to have more felxibility in how we separate the steps.


src/components/MaStepper/MaStepper.spec.js
```diff
@@ -1,39 +1,62 @@
...
 
 function renderComponent(props = {}) {
-  return render(MaStepper, {
+  const utils = render(MaStepper, {
     props,
     slots: { default: slot },
   })
+
+  const content = utils.getByText('Content 1').parentNode
+  const parentNode = utils.getByText('Content 1').parentNode
```
edufarre, line 56: Aren't `content` and `parentNode` the same?

sergigraciaj: True, thx!

icurdu (approved):
LGTM! I will add the maximum and minimum values in the documentation

PR #452, opened 2021-09-10
fix(MaOption): fix indicator shrinking

sergigraciaj:
When we have a long text with more than one line, the indicator shrinks:
image


alex-castelo (approved):
I think all it's OK!

src/components/MaOption/MaOption.css
```diff
@@ -81,6 +81,7 @@
     height: 1rem;
     position: relative;
     transition: border 0.4s, background-color 0.33s;
+    flex-shrink: 0;
```
edufarre, line 84: Shall we add the required prefix for some of the browsers that require it?
https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink#browser_compatibility

edufarre (commented):
Quick question, shouldn't the `indicator` class be `display: flex` in order to use `flex-shrink`?

alex-castelo:
> Quick question, shouldn't the `indicator` class be `display: flex` in order to use `flex-shrink`?

I think indicator it's son of `ma-.radio` selector which does have the `display:flex` property. It should work if I'm not wrong. Maybe I'm missing something.

edufarre (approved):

Samitier (approved):

PR #458, opened 2021-09-29
fix: fix ma option error icon

Samitier:
I was trying to use the errored checkbox in `maitai` but the error icon doesn't load. I don't know if this fixes it but hopefully the svg wil be imported inline and solve the problem.

image


sergigraciaj (approved):

edufarre (approved):

Samitier:
it doesn't work 😂

edufarre:
> it doesn't work 😂

wtf!

Samitier:
I'm going to do it with a `background -mage` like the icon in `ma-text-field`. Seems like webpack inlines those but not `` 🤷‍♂️

PR #460, opened 2021-11-04
feat: add visual feedback when clicking MaButton and MaLink

icurdu:
I've added interactive states to buttons and links components for better user interaction and accessibility. Now, when the user clicks on these components, will have visual feedback that the element was pressed.

Plz, help me with the name of the branch and check if I've messed up with something (I probably did)


afontcu:
I checked it out using the automatic demo site created by Storybook, and it looks great to me! ❤️  great contribution.

As this is an improvement, it should trigger a new Margarita release, a minor version. We're now in the `11.8.6` version, so this should become `11.9.0`. To do so, the commit that ends up in master after merging the Pull Request needs to start with `feat: ` ([explained here](https://github.com/holaluz/margarita#deployment-and-versioning)). The easiest way to do so is to change the Pull Request title, because its title ends up being the commit title. I suggest something like

`feat: add visual feedback when clicking MaButton and MaLink`.

Hit me up if you want to pair on that 😉

afontcu (approved):

edufarre (approved):
🚀 🔥

Issue #461, opened 2021-11-15
Refactor components that not use MaIcon


icurdu:
Refactor margarita components that should use ma-icons but use icons from a separate folder as CSS classes or as img
   - [ ] Alert icons
   - [ ] Checkbox + Radio → Error states
   - [ ] Datagrid order arrows
   - [ ] Select dropdown arrow
   - [ ] TextField → Error State

Samitier:
most of them fixed in https://github.com/holaluz/margarita/pull/471

The ones remaining are a bit more difficult because they have more css, but still doable!

sergigraciaj:
IMO is not worth the effort to use MaIcon in MaOption and MaSelect as we are using the icon to customize checkbox and select HTML elements, and is easier to do in CSS.
What we can consider is to use the svg from the assets project to have one source of truth for svg.

PR #462, opened 2021-11-15
BREAKING CHANGE: New icon system

icurdu:
# The icon system is getting an update! (finally)
And is going to be a big one, with a bit of work to do in everywhere you are using ma-icon in your repo. These changes will improve the visual consistency and standardize the way we use these assets between design and development and for all our users.

Here's the summary of what have changed.
- [x] Modified old icons names and visualization. 
- [x] Add new icons. 
- [x] Modify props to have 3 sizes
    - [x] Remove viewBoxHeight/Width and Width/Height properties
    - [x] Create size property to be a string with 3 values that set the viewbox and the size equally
- [x] Modify props to have defined colors (same as texts)
- [x] Modify components that are affected by icon changes
    - [x] Pagination
    - [x] Text-List (check variant)
    - [x] Textfield with icon story
    - [x] Modal (X to close)

## Next steps after the PR is merged

1. Update Margarita to the last version in the repo
2. Check all ma-icon code insertions and change values (names and props). Here's a cheat sheet to match the old names to the new ones.
icon-names-cheatsheet
4. Check everything looks fine.
5. BONUS TRACK: Check for other icon insertions in your repo that aren't using ma-icon and refactor. Should be pretty easy now ;)

Things to do in the future: [Refactor margarita components that should use ma-icons but use icons from a separate folder as CSS classes or as img](https://github.com/holaluz/margarita/issues/461)


alex-castelo (approved):
Amazing job 👏🏽   I've tested locally the changes and all seems to work properly.  It is an incredible job, many congratulations. I'm sure all the Front teams will thank you for it. 🎉 

I see that there are some small conflicts with `master` branch. If you want we can schedule a small meeting to solve it or we can do it in async.

<a href="https://imgflip.com/i/5uc8r7"></a><div><a href="https://imgflip.com/memegenerator">from Imgflip Meme Generator</a></div>

src/components/MaIcon/MaIcon.vue
```diff
@@ -53,43 +55,22 @@ export default {
...
-
-    /**
-     * Sets icon's view box height
+     * Sets icon's size
+     * @values small, medium, large
      */
-    viewBoxHeight: {
-      type: [Number, String],
-      default: 18,
+    size: {
```
edufarre, line 61: Could we please add a `validator` here and check that the given property values is one of the following? `['small', 'medium', 'large']`

vetur/attributes.json
```diff
@@ -165,64 +125,174 @@
...
-      "LongArrow",
-      "Mail",
-      "Phone",
-      "Plug",
-      "Reassign",
-      "ReassignCircle",
-      "UploadContract",
-      "User",
-      ""
+      "'ArrowRight'",
```
edufarre, line 134: Why do we have all the different names with double quotes `"'POTATO'"`? Is this expected?

edufarre (approved):
Looks good to me! Left a couple of comments. Feel free to ping me or any of the frontenders if you need some help when deploying this or trying to solve the different issues with the master. branch.

src/components/MaIcon/MaIcon.vue
```diff
@@ -109,7 +90,28 @@ export default {
...
+          width: '24',
+          height: '24',
+          viewBox: '0 0 24 24',
+        }
+      }
+      return {
+        width: '32',
+        height: '32',
+        viewBox: '0 0 24 24',
+      }
```
afontcu, line 114: Following the discussion we had over meets, the following code is equivalent and it probably looks a bit better (this is totally subjective, obviously 😄):

```suggestion
      const sizingProperties = {
        small: {
          width: '16',
          height: '16',
          viewBox: '0 0 24 24',
        },
        medium: {
          width: '24',
          height: '24',
          viewBox: '0 0 24 24',
        },
        large: {
          width: '32',
          height: '32',
          viewBox: '0 0 24 24',
        },
      }

      return sizingProperties[this.size]
```

the result is exactly the same (hopefully 😂) but we avoid some repetition and an if/else structure. Instead, we create an object with 3 keys – small, medium, large – with the values for each. Then, depending on the value of `this.size`, we return the values from the appropriate key.

Yet another improvement is to define the "list" of values in the `data` section of the component, and then the `computed` would only be `return this.properties[this.size]`.

In any case, please remember this is simply a refactor and that the behavior is perfectly fine as is 👌  great job

src/components/MaPagination/MaPagination.vue
```diff
@@ -7,7 +7,7 @@
         class="ma-pagination__button ma-pagination__button--backwards"
         @click="onButtonClick(currentPage - 1)"
       >
-        <ma-icon icon="Arrow" width="16" height="16" />
+        <ma-icon icon="ChevronLeft" size="medium" title="Back" />
```
afontcu, line 10: I don't think we need a title here, as the element is nested inside a Button which already has an ARIA label so it will be properly announced:

```suggestion
        <ma-icon icon="ChevronLeft" size="medium" />
```

if we still want to use a title (If I'm missing something), then I'd use the same label:

```suggestion
        <ma-icon icon="ChevronLeft" size="medium" :title="leftButtonAria" />
```

again, I believe this is not needed.

_(same thing with the button on the right)_

src/components/MaIcon/MaIcon.vue
edufarre: Good point! Should this be properly allocated with the rest of our ds tokens? Looks like these design rules could live with the rest of them (even though right now we only use them in the icon, we could end up having `ma-img`?)

afontcu: premature abstraction? 😬  they are only used here and they don't change responsively, so I don't really see the benefit of porting them to an external file (not yet, at least!)

edufarre: Well... if we change the used framework, will these values change? 😬  I wasn't saying we should do these now, but it would be nice to have all the rules that define our Design System apart and isolated from the rest of the framework code.

vetur/attributes.json
icurdu: Nope, not expected!

src/components/MaIcon/MaIcon.vue
afontcu: fair points!

afontcu (approved):
Amazing work mate, looks good to go to me!

If you want we can pair on creating the release, given that it is a breaking change and I believe you've never created one before 😉

src/components/MaIcon/MaIcon.vue
```diff
@@ -109,7 +91,29 @@ export default {
...
+        },
+        large: {
+          width: '32',
+          height: '32',
+          viewBox: '0 0 24 24',
+        },
+      }
+
+      return sizingProperties[this.size]
     },
```
Samitier, line 117: `viewBox` is always the same and `width` is always equal to `height`. Why don't we change this computed to something like:

```js
computedSize() {
  switch(this.size) {
      case 'small': return '16'
      case 'medium': return '24'
  }
  return '32'
},
```
And then we use it in the template like so:

```html
<svg
    xmlns="http://www.w3.org/2000/svg"
    :width="computedSize"
    :height="computedSize"
    viewBox="0 0 24 24"
    class="ma-icon"
    :aria-labelledby="id"
    role="presentation"
  > ...
```
This would simplify things a bit IMHO.

src/components/MaList/MaListMarker.vue
```diff
@@ -1,12 +1,6 @@
 <template>
   <ma-text v-if="type === 'check'" tone="pink">
-    <ma-icon
-      icon="Checkmark"
-      width="16"
-      height="14"
-      view-box-height="14"
-      title="Check icon"
-    />
+    <ma-icon icon="Checkmark" size="small" title="Checkmark Icon" />
```
Samitier, line 3: Maybe the `title` prop is redundant in `MaIcon` and it should be inferred from the name of the icon? For example having a dictionary in `MaIcon` like:

```
ArrowRight -> Icon of an arrow pointing down
Calendar -> Icon of a calendar
...
```

That way all of our icons will be accessible out of the box without having to set a `title` prop explicitly (that we don't use almost anywhere).

We should be aware of internationalization, though. This includes the example avobe too.

src/components/MaIcon/MaIcon.vue
```diff
@@ -109,7 +91,29 @@ export default {
...
   },
 }
```
Samitier, line 119: can we remove the whole:
```
<style scoped>
svg {
  display: inline-block;
  vertical-align: baseline;
}
</style>
```
as well? I don't think it's necessary anymore

Samitier:
Amazing work 🔝  I can't believe the MaIcon will be actually usable.

One think that I think we should bear in mind, we will be loading ALL the svgs when we load margarita if we load the icons as dynamic vue components. Even if we never use them. We are talking of

```
Before:
  dist/margarita.common.js     233.66 KiB  // 38.06 KiB Gzip

After:
  dist/margarita.common.js     328.20 KiB // 50.77 KiB Gzip
```

Maybe we should talk about this, it's pretty substantial TBH

src/components/MaIcon/MaIcon.vue
afontcu: since we might want to extract this outside of the component (see https://github.com/holaluz/margarita/pull/462#discussion_r749987766) I think having it as a simple data structure will make it easier in the future

src/components/MaList/MaListMarker.vue
afontcu: For this line in particular, I think removing `title` is good enough, because the title will be inferred from the component.

For the general picture, I'm not entirely sure having a "Icon of a calendar" description is any helpful. I see two scenarios:
1. One where the icon has a real meaning, and it is the only element (say, a button with an icon to download a contract). In this case we don't want a description for the icon, but rather some additional context (text, ARIA attribute, or the title) that explains what's the element about.
2. Another where the icon is purely stylistic. In this case, I'd argue the icon should be invisible to screen readers, because they only add noise. This is the same scenario with stylistic images, where the suggestion is to drop a `alt=""` attribute.

We might want to iterate this in a future release?

src/components/MaIcon/MaIcon.vue
Samitier: Ah didn't read that. It's ok, I think we can have the data structure, but without the redundant stuff:

```js
const sizingProperties = {
  small: {
    size: '16',
  },
  medium: {
    size: '24',
  },
  large: {
    size: '32',
  },
}
return sizingProperties[this.size]
```

src/components/MaList/MaListMarker.vue
Samitier: Nice! I didn't think of buttons with only an icon inside. Yeah, makes sense if we rethink this a bit, but not in this PR. I would remove this title too then.

src/components/MaIcon/MaIcon.vue
```diff
@@ -53,43 +55,23 @@ export default {
...
+      validator: (val) => ['small', 'medium', 'large'].includes(val),
     },
 
     /**
      * Sets icon's color tone
+     * @values white, red, pink, blue, green, yellow, gray-darker, gray-dark, gray
      */
     tone: {
       type: String,
       default: 'currentColor',
```
sergigraciaj, line 73: `currentColor` is not a valid tone, so we need to set a default valid tone here

Samitier: why not the exact same we have in `ma-text`?

src/components/MaList/MaListMarker.vue
Samitier: in summary:

```suggestion
    <ma-icon icon="Checkmark" size="small" />
```

src/components/MaIcon/MaIcon.vue
icurdu: Go ahead and delete it then!

src/components/MaIcon/MaIcon.vue
icurdu: Makes sense @Samitier!

src/components/MaIcon/MaIcon.vue
sergigraciaj: We decided to go with and empty string as a default, then in the computed we evaluate it to false so we don't fill the 'fill'  attribute : D

Samitier (approved):

PR #463, opened 2021-11-17
[PoC] Accessible disabled ma-text-field

Samitier:
We are having trouble making a form accessible because of the "disabled" ma-text-fields. 

An`<input disabled` does not add a tab stop nor is it focusable. An `<input readonly` adds a tab stop, it's focusable, but it's not writable. [Source](https://medium.com/@izzmo/anti-pattern-using-disabled-or-read-only-for-form-input-e2c24c669f5b).

This problem makes forms like this (address field depends on postal code) unusable, since pressing tab passes from the postal code input directly to the "Tornar enrere" button:

image

I think we should stop using `disabled` in favor of `readonly`. What do you think?

Yes. Maybe having internal dependencies between inputs is the problem here. Maybe someone else has a better example 😂

### DIFF
--- a/src/components/MaTextField/MaTextField.css
+++ b/src/components/MaTextField/MaTextField.css
@@ -35,10 +35,9 @@
   border: 1px solid var(--color-pink-base);
 }
 
-.ma-text-field__input-wrapper--disabled {
+.ma-text-field__input-wrapper--readonly {
   background-color: var(--color-brown-light);
   color: var(--color-gray-base);
-  pointer-events: none;
 }
 
 .ma-text-field__input {
--- a/src/components/MaTextField/MaTextField.spec.js
+++ b/src/components/MaTextField/MaTextField.spec.js
@@ -41,9 +41,9 @@ describe('TextField', () => {
   })
 
   test('renders a disabled input', () => {
-    const { input } = renderComponent({ disabled: true })
+    const { input } = renderComponent({ readonly: true })
 
-    expect(input).toBeDisabled()
+    expect(input).toHaveAttribute('readonly', 'readonly')
   })
 
   test('renders custom id', () => {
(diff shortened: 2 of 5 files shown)


afontcu:
Given that:

1. `disabled` and `readonly` are different things, and one cannot be confused with the other (as you pointed out, good catch).
2. This PR is a breaking change.

Would it be simpler to add styles for a `readonly` input element? I mean, you can already define a readonly input element:

```vue
<!-- this already works! it is pure HTML. if it doesn't, then it's a bug -->
<ma-text-field readonly />
```

but we might be missing some styles. What about adding them letting `disabled` keep working? There are some legit use cases for it, and instead of a breaking change this would mean releasing a minor version.

Samitier:
Yes it's only for the styles! It's not that easy to add the styles because we need `readonly` to be a prop. The element that has the css for `disabled` is the container of the `input`, therefore we need to be able to set a custom css class there.

Since we saw we have to maintain both props `disabled` and `readonly`, we deleted `disabled` 😈.  but yeah it's a breaking change. I don't know what should we do. We can patch it and add the css directly to the `input` but this might break other stuff of ma-text-field.

Or if no one is interested in this, we simply put the `::v-deep` in `maitai` and call it a day 😂

afontcu:
> It's not that easy to add the styles because we need `readonly` to be a prop. The element that has the css for `disabled` is the container of the `input`, therefore we need to be able to set a custom css class there

yep, makes total sense!

> Since we saw we have to maintain both props disabled and readonly, we deleted disabled 😈. but yeah it's a breaking change. I don't know what should we do.

Since the maintenance surface is really small, I vote for keeping them both, because they are just a couple of CSS classes. We'll avoid the hassle of a breaking change + keep the `disabled` prop that might come in handy.

---

It is true, though, that we might want to check what are the current usages of `disabled`, and assess whether if `readonly` is a better solution. Thank you for bringing this up!

PR #466, opened 2021-12-02
chore: Add semantic release checker

sergigraciaj:
We just need to check for semantic release prefix in the title as we squash the commits, but as this is the PR to add the configuration, the prefix has to be in the commit message too in order to not fail the check.

Edit (thx to Alex):
Here the [docs](https://github.com/zeke/semantic-pull-requests)


alex-castelo (approved):
I've finally found the [docs](https://github.com/zeke/semantic-pull-requests) related to this. Seems good to me!

Issue #467, opened 2021-12-13
Remove arialabelledby in ma-icon


carlotamolintr:
Hello all,

Following the indications giving in maitai accessibility report, we need to remove the arialabelledby in ma-icon component. 

See the recommendation here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role

![Captura de pantalla 2021-12-13 a las 9 58 17](https://user-images.githubusercontent.com/51370533/145782081-c13c6855-8f29-4194-93d1-8e7783dee5a7.png)

Samitier:
fixed in https://github.com/holaluz/margarita/pull/471

PR #468, opened 2021-12-13
fix: Remove arialabelledby

carlotamolintr:
Related with: #467 

Following the indications giving in maitai accessibility report, we need to remove the arialabelledby in ma-icon component.

See the recommendation here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role

and the report here
[Auditoria Alta-holaluz.pdf](https://github.com/holaluz/margarita/files/7702415/Auditoria.Alta-holaluz.pdf)

Please, let me know if you have any comments. :)


afontcu (approved):

carlotamolintr:
We won't this anymore as the #471 has already removed it.

PR #469, opened 2021-12-23
feat: Add MaExpansionPanel component

edufarre:
aka: Accordion


src/components/MaExpansionPanel/MaExpansionPanel.vue
```diff
@@ -0,0 +1,92 @@
...
+      return {
+        width: this.labelWidth ? `${this.labelWidth}px` : 'fit-content',
+      }
+    },
+  },
+
+  watch: {
+    initiallyExpanded(newValue) {
+      this.isContentExpanded = newValue
+    },
```
afontcu, line 73: Wha's the purpose of this watcher? Is this to make sure we toggle visibility if prop changes its value? If that's the case, I think that's an unexpected behavior: the prop is specifically called `initiallyExpanded`, so changing the prop should not change behavior once the componentis initially rendered, right?

src/components/MaExpansionPanel/MaExpansionPanel.vue
```diff
@@ -0,0 +1,92 @@
...
+  created() {
+    this.isContentExpanded = this.initiallyExpanded
+  },
+
+  methods: {
+    toggle() {
+      this.isContentExpanded = !this.isContentExpanded
+      this.$emit('toggled', this.isContentExpanded)
+
+      if (this.$parent?.$options?.name === 'MaExpansionPanelGroup')
```
afontcu, line 85: Where's "MaExpansionPanelGroup" coming from?

edufarre: It will be incoming in another (next) PR :)

afontcu: oh, ok – it should be gone for now, then?

src/components/MaExpansionPanel/MaExpansionPanel.vue
edufarre: Then I guess the problem is the naming of the prop. I would go for something like `isExpanded`.

src/components/MaExpansionPanel/MaExpansionPanel.css
```diff
@@ -0,0 +1,31 @@
...
+  align-items: center;
+  cursor: pointer;
+}
+
+button:focus:not(:focus-visible) {
+  outline: none;
+}
+
+.chevron {
+  transition: all 0.3s;
```
Samitier, line 16:
```suggestion
  transition: transform 0.3s;
```
?

src/components/MaExpansionPanel/MaExpansionPanel.vue
```diff
@@ -0,0 +1,89 @@
...
+  name: 'MaExpansionPanel',
+
+  props: {
+    /**
+     * Sets if content should be displayed on initial render
+     */
+    initiallyExpanded: {
+      type: Boolean,
+      default: false,
+    },
```
Samitier, line 45: can't we merge this prop `initiallyExpanded ` with `isContentExpanded `? Isn't it redundant having two? We can even make it a `v-model` so that we can check when the user opens the expansion panel (for example, for tracking).

src/components/MaExpansionPanel/MaExpansionPanel.vue
Samitier: this would not matter if we did this: https://github.com/holaluz/margarita/pull/469/files#r775018260

afontcu: yep, agreed

src/components/MaExpansionPanel/MaExpansionPanel.vue
edufarre: yes! However, that means we are gonna need a v-model for each expansion panel we want to use, instead of leaving its responsibility to the component itself. The `initiallyExpanded` was more like a helper function, that allow us to set if we would like to display or not the content when it is initially rendered. And the `emit` was added a few days ago, in order to trigger a GA/DD event when the user clicks on it.

The intention never was to handle the component state from its parent, in my mind it did sound better to allow to handle it by itself. Moreover, the `watch` function is because we desired to change its behaviour when being on mobile screens, thus avoiding lots of scroll because of this element.

I believe maybe we should discuss the desired approach instead of going back and forth with it, and just implementing it once. I basically ported what we currently have in negroni to the project, as we agreed into yesterday's meeting, but I am more than happy to iterate this and try to find a case that suits us all.

src/components/MaExpansionPanel/MaExpansionPanel.vue
edufarre: I believe it depends of the approach we would like to tackle

Samitier:
hello! We now have expansion panels in maitai and they look like the ones in dry-martini. Should we add them here too?
image

edufarre:
> hello! We now have expansion panels in maitai and they look like the ones in dry-martini. Should we add them here too? image

As I said when we were discussing the expansion-panel component. I would start for the simple case (even tho we are still discussing if we should use the `v-model` or not, never got any other reply to my last comment), and then add in future iterations the properties we would like to add on top of the basic case.

This being said, I guess you are talking about the 'card' style, right? Is not as easy as wrapping this component into a `ma-card` one? 🤔

alex-castelo:
> As I said when we were discussing the expansion-panel component. I would start for the simple case (even tho we are still discussing if we should use the v-model or not, never got any other reply to my last comment), and then add in future iterations the properties we would like to add on top of the basic case.
I totally forgot this. I think this is a good approach. Let's create a "simple" version of these and iterate it depending on our needs. What do you think about it?

Issue #464, opened 2021-11-26
Load icons dynamically


icurdu:
One thing that I think we should bear in mind, we will be loading ALL the svgs when we load margarita if we load the icons as dynamic vue components. Even if we never use them. We are talking of

```
Before:
  dist/margarita.common.js     233.66 KiB  // 38.06 KiB Gzip

After:
  dist/margarita.common.js     328.20 KiB // 50.77 KiB Gzip
```

Maybe we should talk about this, it's pretty substantial TBH

_Originally posted by @Samitier in https://github.com/holaluz/margarita/pull/462#issuecomment-971598635_

Samitier:
fixed in https://github.com/holaluz/margarita/pull/471

PR #471, opened 2021-12-28
fix: load MaIcons as SVG from assets.holaluz.com

Samitier:
Solves https://github.com/holaluz/margarita/issues/464 by loading the icons from https://assets.holaluz.com instead of vue components. In order to be able to change the color of the SVGs by code, we need to use `mask-image`, which comes with a couple of caveats written below.

## Does it solve the problem

Yes! Also the sizes are smaller than when we added the icons 😂 

```
  dist/margarita.umd.min.js    51.37 KiB  //  15.49 KiB Gzip
  dist/margarita.umd.js        190.14 KiB  //  30.73 KiB Gzip
  dist/margarita.common.js     189.67 KiB //  30.58 KiB Gzip
  dist/margarita.css           32.90 KiB   //    6.41 KiB Gzip
```

## Why mask-image?

We are loading icons with mask-image because we need to be able to change the color of the loaded svg. Loading the icons with an `img` tag will make them unable to change color via css (unless we hack it away with `filter`).

We could also use an `object` tag, but we will be unable to change color of the icons if the domain is different than assets.holaluz.com (for example, localhost). That has to do with iframe security, we can't modify the contents of an iframe from a different domain.

And lastly, we can load the svg as plain text and put it into vue using `v-html`. It will work like it worked before, but we open the possiblity of xss attacks and also we lose the ability to be able to cache the icons in assets.holaluz.com.

So mask-image imho is the best option.

## Caveats

-  The browser usage of `maskImage` is a bit weird, but I tried all main browser's old versions and it works fine! 
-  For accessibility, I don't know if we should have a `title` prop or we should completely ignore icons with an `aria-hidden="true"`. I put them both 😂 
- Had to manually specify a tone to icons that have a default color (like the ones in ma-alert). This is kinda good because now all icons behave the same way, but we will need to be careful that everything still looks the same (breaking change?).
- We are rendering icons in a completely different way than we used to do and that can come with unexpected results. Sooo... is it a breaking change?


src/components/MaIcon/MaIcon.vue
```diff
@@ -109,5 +91,18 @@ export default {
...
+
+  watch: {
+    tone() {
+      this.changeColorOfSvg()
+    },
+  },
+
+  methods: {
+    changeColorOfSvg() {
+      this.$refs.iconElement.contentDocument.children[0].style.fill =
```
alex-castelo, line 103: Nice PR, I think the script used to create all icons is a good idea 👍🏽  . I have doubts about this line? Where does come from this `contentDocument.children[0]` ? We use this to get the parent color?

Samitier: hahahah it's precisely this what I'm referring to when I say:

> Also the code that sets the color of the icon is a bit hacky.

Basically, we are loading the svg as an `<object>` tag, which is (kind of) an iframe. In order to change the color of the svg, we need to access the document it is loading inside of itself, and set the css `fill: color` to the `<svg>`.

So imagine something like:

```html
<object type="image/svg+xml">
  #document
    <svg (the icon) /></svg>
</object>
```

We only have the reference of the `<object>` tag (`this.$refs.iconElement`), so to access the svg we need to do:

```
this.$refs.iconElement.contentDocument.children[0].style.fill = color
    (object tag)        (html document) (svg tag)  (setting fill css in the svg)
```

And all this only when the icon has loaded. That's why this function is called on the `@load` of the `object` tag.

alex-castelo: Wow! Amazing. Thanks for the clarification Blai

Samitier: I ended up doing something else. What I did avobe won't work if the icon comes from a different domain as the one you try to load it because of security reasons.

Since the icons come from assets.holaluz.com and we develop in localhost, it will give an error.

src/components/MaIcon/MaIcon.vue
```diff
@@ -1,27 +1,19 @@
...
-  </svg>
+    :style="{
+      backgroundColor,
+      width: sizeInPx,
+      height: sizeInPx,
+      maskImage: `url(https://assets.holaluz.com/icons/${icon}.svg)`,
+    }"
+    :title="icon"
+    aria-hidden="true"
+  />
```
Samitier, line 12: we could also do something more similar than what it was before, loading the svg with a fetch request:
```vue
  <svg
    :width="sizeInPx"
    :height="sizeInPx"
    aria-hidden="true"
    :fill="fill"
    v-html="iconSvg"
  />

async fetchIcon() {
  this.iconSvg = await (
    await fetch(`https://assets.holaluz.com/icons/${this.icon}.svg`)
   ).text()
    this.iconSvg = this.iconSvg.replace(/^.*/g, '').replace(/.*$/g, '') // removes svg tag to leave only the body
},
```

carlotamolintr:
Hi Blai,

As we agreed, could you please remove the `aria-labelledby="id"` in this PR? Related with #468 

thank you

Samitier:
> title

done!

PR #473, opened 2022-01-13
feat: port margarita to TypeScript

afontcu:
There are several things that need to be figured out, but it's a start. Todo:

- [x] install TS, port vue cli to v5, update required deps
- [x] `npm start` should work
- [x] `npm test` should work
- [x] `npm run lint` should work
- [x] `npm run build` should work. I haven't check if the build is correct, though
- [x] `npm run format` should work
- [x] port `tokens` to TS

~port `markdown` directive to TS~ (to be handled in another PR)
~port stories to TS~ (to be handled in another PR)

I have only ported some `.vue` files as a demo – check out MaIcon for instance.

Feel free to ask anything unclear or with room for improvement! Also, any help and double-checking that everything is still working will be much appreciated.


src/index.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+test('exposes an install function', () => {
+  expect(defaultExport).toHaveProperty('install')
+})
+
+// TODO: This test currently doesn't work because components' shape is different
+// when they are wrapped in Vue.extend(), so we can't get a list of them by simply
+// filtering by m.props on line 37. We need to find a way to query extended and
+// non-extended components, or extend them all.
+test.skip('installs Margarita components', () => {
```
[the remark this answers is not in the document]
alex-castelo: I want to try 🚀 . Just to clarify, the point is that the components defined with the new syntax `Vue-extend()` have a different shape from those defined in the old way. My question is, what do we need to do?

- Create a function that facilitates the assertion in the tests and allows us to compare the 2 types of final objects.
- Pass all the components to the new Vue.extend syntax.

Maybe my question is a bit stupid. 😅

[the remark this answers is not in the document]
afontcu: `Vue.extend()` is the only way for Vue/TS/the IDE to know that a plain object (such as the one we define in a component) is actually intended to be a component: https://vuejs.org/v2/guide/typescript.html#Basic-Usage

> Pass all the components to the new Vue.extend syntax.
that would be cool, but requires some work.

> Create a function that facilitates the assertion in the tests and allows us to compare the 2 types of final objects.
right we're only checking if the object has a `prop` key, and determine that "a component is a component" given that. We'd need to find some similar assertion to check for the other components 👍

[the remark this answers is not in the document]
alex-castelo: Maybe I have found a way to do it. See if it suits you for the moment or if you have an alternative.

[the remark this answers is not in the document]
afontcu: LGTM! We could maybe do it in a single loop, but given that the array is like ~20 items long… that's more than good enough 👌👌

Samitier:
Hello, thanks for getting things started! 💪 🚀 

I only have one suggestion:

Maaaaybe since we are changing the whole pipeline and rewriting everything anyways, we could consider doing this? https://github.com/holaluz/kalimotxo/pull/97 (using `script setup` and volar)

As I've said before, typing components in "vanilla" Vue2 IMHO actually worsens the developer experience. But what if we could actually have 3 wins at the same time:
  - Using Typescript and actually make coding components easier (that's a bit subjective but at least vscode won't go insane with the type checking).
  - Achieve 100% typescript coverage. And I mean 100% (even the templates and the styles will be checked).
  - Prepare our way to Vue 3. If we use `script setup`, we can practically use the code as-is except by the imports.

I know it sounds like a lot of work, but honestly typing the whole Margarita already is, maybe doing the hard work only once makes our life easier in the future.

afontcu:
hey!

> Maaaaybe since we are changing the whole pipeline and rewriting everything anyways

I didn't expect this PR to rewrite everything, but simply set up the TS pipeline – actually wrapping components in `Vue.extend()` is a nice to have, but not required at all. 


> typing components in "vanilla" Vue2 IMHO actually worsens the developer experience

I don't fully agree – It is true that Vue 3 DX with TS is top notch, but Vue 2 with `Vue.extend()` and template interpolation from Vetur is good enough and definitely better than using plain JS in terms of build-time errors, code completion, and so on.


> Prepare our way to Vue 3. If we use script setup, we can practically use the code as-is except by the imports.

Notice that Vue 3 does not require `<script setup>`. Porting current components to Vue 3 means fixing the breaking changes, but not updating to `<script setup>`. It is a nice addition, but it is by no means required.

Samitier:
> I didn't expect this PR to rewrite everything, but simply set up the TS pipeline – actually wrapping components in `Vue.extend()` is a nice to have, but not required at all.

I mean practically rewriting the entire component when moving it to TS. Note that in that Kalimotxo POC, only the components with`<script setup` will actually use the new syntax, the rest work normally. To me, what is hurts is "wasting" time rewriting a component in TS for vue 2 when we can move it to vue 3 syntax and have better TS support for free. But the incremental update can be done regardless.

> I don't fully agree – It is true that Vue 3 DX with TS is top notch, but Vue 2 with `Vue.extend()` and template interpolation from Vetur is good enough and definitely better than using plain JS in terms of build-time errors, code completion, and so on.

Yes, but again better TS support can be achieved right now, with a syntax that it's future-proof. Also, we have intelisense/typechecking in templates, reference checking in styles, and proper type inference everywhere.
 
> Notice that Vue 3 does not require `<script setup>`. Porting current components to Vue 3 means fixing the breaking changes, but not updating to `<script setup>`. It is a nice addition, but it is by no means required.

Correct me if I'm wrong, but doesn't porting them to `script setup` already fix the breaking changes that they could have? Of course we will still need to update the vue imports so that they import from `vue` instead of `'@vue/composition-api'`, but the rest will be already done, no? It can also be a nice exercise for learning to use the new syntax for everyone involved, and it can be done incrementally whenever we feel like doing it. There will be limitations of course, like top level awaits and stuff like that but IMHO we will never use them in margarita anyways.

src/components/MaButton/MaButton.vue
```diff
@@ -93,23 +106,23 @@ If you're rendering a link with a button using tag="a", please consider switchin
   },
 
   methods: {
-    handleClick(event) {
+    handleClick(event: Event) {
```
carlotamolintr, line 109: Is this type "Event" from html own tags?

alex-castelo: I think so. It was autosuggestion from VS Code, and I think it is related to the types described for this library. https://github.com/microsoft/TypeScript/blob/main/lib/lib.dom.d.ts . As you said it, is a way to type HTML DOM events.

Maybe it's not the more accurate way to do it but reading the description seems it matches our case.

src/components/MaButton/MaButton.vue
```diff
@@ -93,23 +106,23 @@ If you're rendering a link with a button using tag="a", please consider switchin
...
       /**
        * Click event
        * @event click
        */
       this.$emit('click', event)
     },
 
-    getLoadingStyle() {
+    getLoadingStyle(): Record<string, string> {
```
carlotamolintr, line 117: Why we use here Record<Keys, Type>? I mean, could not only be: getLoadingStyle(): string to type it?

alex-castelo: Very good question. Actually, I think I saw the example in another part of the code and thought it worked for our case.

Reading the documentation for Record<key, type> I see that maybe it has more value for composing complex types. For example if my Key instead of String, was of a Margin type created by us, maybe it would make more sense.

In this case, the return of this computed property is not a String, but an object with a Key of type String and a value also of type String.

I don't know if this is understood a little better :)

.stylelintignore
```diff
@@ -3,3 +3,5 @@
 *.eot
 *.ttf
 *.json
+*.stories.js
+
```
carlotamolintr, line 7: At least, I just ignored the stories files to avoid the cssSyntax error. I hope it will find like that. :)

carlotamolintr:
All the following commands working: 

- [ ] npm start
- [ ] npm test
- [ ] npm run lint
- [ ] npm run build
- [ ] npm run format

Samitier (approved):

PR #478, opened 2022-01-25
fix: ma-tabs reactivity issue

### LINKED ISSUE #474: Reactivity breaks inside a ma-tab
Super weird bug in `ma-tabs`. Seems like reactivity is breaking inside the slot of a ma-tab. For example:

```vue
<template>
  <ma-tabs v-model="selectedTab">
    <ma-tab title="A" name="A">
      {{ aComputedProp }}
    </ma-tab>
    <ma-tab title="B" name="B">
      {{ aComputedProp }}
    </ma-tab>
  </ma-tabs>
</template>

<script>
export default {
    data() {
        return {
             selectedTab: 'A'
        }
    },
    computed: {
        aComputedProp() {
           return this.selectedTab === 'A' ? 'tab A selected' : 'tab B selected'
        }
    }
}
</script>
```

The computed works properly and updates properly in the template if set outside the tab body. When inside the tab body, the computed stays as `tab A selected` forever.

Something is breaking reactivity inside the slots of the ma-tab, probably that render function we have in ma-tabs is not updating when it should.

Samitier:
Fixes #474 .

If we use computed that return vnodes, we will lose all reactive behavior (for example computed or prop changes). To solve this, we simply call these computed as methods.


afontcu (commented):
LGTM but I'd suggest adding a test to make sure that it is working now and it doesn't break in a near future 😄

Samitier:
> LGTM but I'd suggest adding a test to make sure that it is working now and it doesn't break in a near future 😄

Done! @afontcu

carlotamolintr (approved):

afontcu (approved):
👌

PR #481, opened 2022-02-09
fix: error icon displayed when input has suffix

sergigraciaj:
As proposed in [Watt](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=4215%3A7819), MaTextField shouldn't display the error icon when it has suffix.


Samitier (approved):
🙇 I definitely broke this in https://github.com/holaluz/margarita/pull/471, sorry! Great fix 😄

PR #482, opened 2022-02-10
feat: publish tokens without publishing entire source

Samitier:
We are publishing the entire margarita source code to npm. I think we should only publish what we need from src, directly as dist, no? I think its only the `tokens.css` file.

(I have no idea if this works, we should publish a beta first)


alex-castelo (approved):
🔝  Not sure if we need an appoval to publish the beta version. But , let's try!

edufarre:
we could always use [npm link](https://docs.npmjs.com/cli/v8/commands/npm-link) instead of publishing a `beta` version, right?

PR #483, opened 2022-02-10
fix: modify ma-tabs

sergigraciaj:
- Modify hover effect to follow [Watt proposal](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=5100%3A7891).
- When tabs are in a form and a tab is clicked, the form is sent. So we add a preventDefault to avoid that.


src/components/MaTabs/MaTabs.vue
```diff
@@ -32,24 +33,32 @@ export default {
...
   render() {
     const selectedTab = this.getSelectedTab()
     const tabHeaders = this.getTabs().map((child) => {
       const { title, icon, pill, name } = child.componentOptions.propsData
       const isSelected = selectedTab === child
-      const tone = isSelected ? 'pink' : 'gray-dark'
-      const pillTone = isSelected ? 'pink' : 'dark'
+      const isHovered = this.hoveredTabName === name
+      const tone = isSelected ? 'pink' : isHovered ? 'pink' : 'gray-dark'
+      const pillTone = isSelected ? 'pink' : isHovered ? 'pink' : 'dark'
```
Samitier, line 53:
```suggestion
      const tone = isSelected || isHovered ? 'pink' : 'gray-dark'
      const pillTone = isSelected || isHovered ? 'pink' : 'dark'
```

src/components/MaTabs/MaTabs.vue
```diff
@@ -32,24 +33,32 @@ export default {
         (t) => t.componentOptions.propsData.name === this.selectedTabName
       )
     },
-    setSelectedTabName(name) {
+    setSelectedTabName(name, e) {
+      if (e) e.preventDefault()
```
Samitier, line 37: why? Just curious, not that there's anything wrong with it 😂

src/components/MaTabs/MaTabs.vue
```diff
@@ -32,24 +33,32 @@ export default {
...
       this.selectedTabName = name
       this.$emit('input', name)
     },
+    setHoverTabName(name) {
```
Samitier, line 41: Just for consistency with the var name maybe?

```suggestion
    setHoveredTabName(name) {
```

src/components/MaTabs/MaTabs.spec.js
```diff
@@ -1,9 +1,14 @@
...
-import { render, waitFor } from '@margarita/margarita-test-utils'
+import { render, waitFor, fireEvent } from '@margarita/margarita-test-utils'
 import userEvent from '@testing-library/user-event'
 import MaTabs from '.'
 import MaTab from '../MaTab'
 
+const mouseenter = new MouseEvent('mouseenter', {
+  bubbles: false,
+  cancelable: false,
+})
```
Samitier, line 10: this is only used in one test. Maybe we can define it inside the test instead of "globally" in the spec file?

src/components/MaTabs/MaTabs.spec.js
```diff
@@ -88,6 +93,27 @@ describe('Tabs', () => {
...
+      },
+    })
+
+    await fireEvent(getByText('Header Tab1'), mouseenter)
+
+    expect(getByTitle('Gas')).toHaveStyle({
+      'background-color': 'rgb(230, 0, 125)',
+    })
+    expect(getByText('Header Tab1')).toHaveStyle({
+      color: 'rgb(230, 0, 125)',
```
Samitier, line 113: does it work if we import the colors from `token/color.js`? That way we don't hardcode the color in the test:

```
import color from '@/tokens/color'

expect(getByText('Header Tab1')).toHaveStyle({
      color: color.pink.base,
```

No idea if something like that would work

Samitier (approved):

src/components/MaTabs/MaTabs.spec.js
sergigraciaj: Color tokens are hexadecimal while the rendered one is rgb : /

Samitier: hahah 💩

src/components/MaTabs/MaTabs.vue
Samitier: hahahah oh god. Well, thank god you found that

PR #501, opened 2022-03-09
feat: Add style for links inside MaText

sergigraciaj:
When we need to use links inside text, it's a mess to combine _MaTexts_ and _MaLinks_, and though _MaText_ has the span prop, it is not really inline as it has `display: table` in the `:before` and `:after`.
So we decided to style the links inside the _MaText_ in the _MaText_ itself.
More info in [this slack thread](https://holaluz.slack.com/archives/C019DLZTBU6/p1646821991705009)


alex-castelo (approved):
I think this helps to unify the links inside text components and I saw your conversation with Iban, so, thanks for doing it and let's move on 💪🏽 

Also I understand that If we want a custom link we can use a `MaLink` nested component inside a `MaText` with their own custom styles.

PR #503, opened 2022-03-18
fix: Change line height in MaHeading

carlotamolintr:
BREAKING CHANGE: Change line-height in MaHeading. Instead of 1.5 will be 1.1


src/tokens/headings.ts
```diff
@@ -63,7 +63,7 @@ export const headings = {
     },
     xlarge: {
       'font-size': '4rem',
-      'line-height': 1.5,
+      'line-height': 1.1,
       'top-crop': '-0.405em;',
       'bottom-crop': '-0.42em',
     },
```
Samitier, line 69: we have to touch the top crop and bottom crop so that the text gets aligned correctly. No idea how they calculated the previous value though 😂

alex-castelo (commented):
LGTM! But I would like to understand why the 1.5" size was established before. 👍🏽

src/tokens/headings.ts
icurdu: It was calculated manually using the chrome inspector and playing with the numbers until the background was aligned to the taller letter (M or A) and the baseline. There is no formula, just good eyes

icurdu (approved):
Looks good! The previous value was badly documented in Watt Library, and we didn't see the mistake because the preview uses one line text, so we couldn't see the behavior.

edufarre (approved):
@carlotamolintr can we move forward and merge this PR?

PR #519, opened 2022-04-12
feat: Add custom media file

xevops:
This is a first approach to generate a custom-media.css file from breakpoint tokens for reusability and consistency purposes across projects.

After struggling with StyleDictionary, any ideas or suggestions are wellcome


sergigraciaj (approved):

src/css/custom-media.css
```diff
@@ -0,0 +1,8 @@
+/**
+ * Do not edit directly
+ * Generated on Tue, 12 Apr 2022 09:21:57 GMT
+ */
+
+@custom-media --ma-mobile (max-width: 899px);
```
xevops, line 6: Is mobile breakpoint necessary? Probably not, but IMHO I think it's useful in some scenarios where just needs to apply some styles in mobile version.
Opinions?

.style-dictionary/filesGenerators.ts
```diff
@@ -0,0 +1,69 @@
...
+    name: 'custom-media',
+    formatter: function ({ dictionary, platform, options, file }) {
+      return fileHeader({file}) + dictionary.allTokens.map(token => token.value).join(`\n`)
+    }
+  })
+  .registerTransform({
+    name: 'custom-media',
+    type: 'value',
+    transformer: (prop) => {
+      if (prop.name === breakpointsEnum.tablet) {
```
xevops, line 48: Any suggestions about how to deal with this situation?  If we want to add a mobile custom media this is the way I found to achieve it with the current token definitions

src/css/custom-media.css
alex-castelo: The leanest approach would tell you not to do it if you haven't found your use case. On the other hand, if you already imagine scenarios where you will use it, it's fine for me to have it. 👍🏽

alex-castelo (approved):
Nice approach! I am only concerned about this [PR](https://github.com/holaluz/margarita/pull/517) that Edu is working on. I think will have conflicts for sure... But from what I've seen, maybe with a little overtaking and a bit of tact, there won't be a problem.

Having said that. Good work, LGTM

src/css/custom-media.css
xevops: Yes, in fact we're using the --mobile breakpoint in some scenarios right now. For example, in the mobile version of the new solar landing there's some places where it appears a border-top just in mobile version (not in desktop). The alternative would be do something like this:

```
.class-name {
    border-top: bla bla bla...

    @media (--desktop) {
        border-top: none;
    }
}
```

With the mobile custom media the result would be less verbose
```
.class-name {
    @media (--mobile) {
        border-top: bla bla bla;
    }
}
```

With both approaches is possible to do the job, the decision it's just a crew agreement about which one is preferred

xevops:
> 

Yes totally agree. In fact I think would be better to wait for Edu before merge this PR to avoid or at least reduce the headaches with possible code conflicts

src/css/custom-media.css
carlotamolintr: For me will be easier to have it! Because dry-martini is already working with a mobile breakpoint...So, if  margarita and dry breakpoints are living together I guess is easier if they all work the same.

edufarre: I believe the mobile breakpoint should not be here, as we should always implement CSS minding the mobile first design

xevops: My doubts about this breakpoint came from the mobile first principle, but also I think it's possible to develop CSS following Mobile First concepts keeping the breakpoint for those escenarios where it is useful to have it. IMHO it's a developer responsability to apply properly mobile first and still will be possible to develop CSS without following mobile first althought this breakpoint will not be present on margarita

.style-dictionary/filesGenerators.ts
edufarre: I would not tackle the mobile breakpoint, thus always 'forcing' to implement CSS minding the Mobile First development. This being said, I believe this is going to make it way easier avoiding the use of the `if` condition

package.json
```diff
@@ -21,7 +21,7 @@
     "prepare": "node check-node.js",
     "build": "vue-cli-service build --target lib --name margarita src/index.ts",
     "lint": "eslint 'src/**/*.{js,vue}' && stylelint 'src/**/*.css'",
-    "prepublishOnly": "cp src/css/tokens.css dist/tokens.css",
+    "prepublishOnly": "cp src/css/tokens.css src/css/custom-media.css dist/",
```
edufarre, line 24: I do not really understand why are we doing this? 
The approach we followed for the `tokens` case was bearing in mind they would be used without installing margarita, thus having to access `dist/tokens` in the desired project in order to load them

.style-dictionary/filesGenerators.ts
```diff
@@ -0,0 +1,69 @@
...
+    platforms: {
+      css: {
+        transforms: [
+          'attribute/cti',
+          'name/dashed',
+          'time/seconds',
+          'content/icon',
+          'size/rem',
+          'color/css',
+        ],
```
edufarre, line 23:
```suggestion
      transformGroup: 'css',
```

https://amzn.github.io/style-dictionary/#/transform_groups?id=css

.style-dictionary/filesGenerators.ts
```diff
@@ -0,0 +1,69 @@
...
+import { breakpointsEnum } from '../src/tokens/breakpoint'
+const { fileHeader } = StyleDictionary.formatHelpers;
+
+export function generateTokensFile() {
+  StyleDictionary
+  .registerTransform({
+    name: 'name/dashed',
+    type: 'name',
+    transformer: (prop) => prop.path.join('-'),
+  })
```
edufarre, line 11: If we use the `transformGroup` instead of the list of `transforms` there is no need to register this one, as it is already included in the group 😄 
See https://amzn.github.io/style-dictionary/#/transform_groups?id=css

package.json
xevops: This script do a copy of tokens.css and custom-media.css files to the dist /folder in the same way it did before adding the custom-medias. From a third project it will be possible to load the files from /dist/tokens.css (as it happens now) and /dist/custom-medias.css

edufarre: Sweet, gotcha!

However, thinking out loud, what do you think about having them with the `tokens.css` file? IN mi mind I dunno why I still see them as tokens (media related, but still tokens). Anyway, we do not have to do it in this PR 👍🏽

xevops: I see them as differents files, so you can import tokens.css file without importing custom-media.css or viceversa depending on project needs. As far as I know tokens file it is used in some specific situations, so it makes sense for me keeping them as separate files

PR #523, opened 2022-04-21
fix: add declarations file

sergigraciaj:
Add TS declarations file in order to work as a plugin in projects with Typescript.
Not sure why we had a margarita declaration file with Layout and Breakpoint types.
Our package points to index.d.ts https://github.com/holaluz/margarita/blob/2bda992f94550e1788577e6ccacf24dab618426d/package.json#L11


alex-castelo (commented):
I'm not sur if I understand. This new file is the one that our npm package points as a `TS declaration file` , right?

sergigraciaj:
That's it : )

alex-castelo (approved):
🔝

PR #524, opened 2022-04-22
refactor: Port ma list to TS

alex-castelo:
## Attention 🚨
It is the first functional component that we migrated. If you think the typed ones work, it can be a good guide on how to migrate the next ones.

PD: I don't know why the test was failing and I had to change it.


sergigraciaj (approved):
Lugutumi 👍

carlotamolintr (approved):

PR #528, opened 2022-04-27
refactor: port MaTextField to TS

sergigraciaj:


src/components/MaTextField/MaTextField.vue
```diff
@@ -149,22 +151,23 @@ export default {
...
+    inputWrapperClasses(): Record<string, boolean> {
       return {
         'ma-text-field__input-wrapper--disabled': this.disabled,
         'ma-text-field__input--error': this.hasError,
       }
     },
 
-    inputListeners() {
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    inputListeners(): Record<string, Function | Function[]> {
```
sergigraciaj, line 168: Not sure this is the best way, from one side the eslint docs for this rule states 
image
But from the other side, vue types this like
image

So maybe we can make the eslint exception here.
Anyways if someone as a suggestion here is welcome : D

src/components/MaTextField/MaTextField.vue
```diff
@@ -149,22 +151,23 @@ export default {
...
   computed: {
-    labelClasses() {
+    labelClasses(): Record<string, string> {
       return {
         'visually-hidden': this.$attrs['aria-label'],
       }
     },
 
-    inputWrapperClasses() {
+    inputWrapperClasses(): Record<string, boolean> {
```
edufarre, line 160: Not sure about this one either

src/components/MaTextField/MaTextField.vue
```diff
@@ -149,22 +151,23 @@ export default {
     }
   },
   computed: {
-    labelClasses() {
+    labelClasses(): Record<string, string> {
```
edufarre, line 154: Not sure about this return type tho

src/components/MaTextField/MaTextField.vue
sergigraciaj: the props we use as value are both boolean, why are u not sure?

src/components/MaTextField/MaTextField.vue
sergigraciaj: `Vue.$attrs` as a string value
image

edufarre: But the computed class is not the same type right? the actual return type would be:
```ts
type LabelClasses = { 
  'visually-hidden': boolean
}
```

Or am I missing something??

src/components/MaTextField/MaTextField.vue
edufarre: It is kinda the same doubt I do have with the other computed property. In my mind, the return type of this computed would be:
```ts
type InputClasses = {
  'ma-text-field__input-wrapper--disabled': boolean,
  'ma-text-field__input--error': boolean,
}
```

And, maybe something even cooler, like having a generic ComputedClass type: 
```ts
type ComputedClasses = {
  [className: string]: boolean
}
```

Thus doing something like:
```ts
type GenericClass = {
  [key: string]: boolean
}

const computedClasses: GenericClass = {
  hidden: true,
  bold: false,
}
```

src/components/MaTextField/MaTextField.vue
sergigraciaj: image

I guess is a truthy string.

edufarre: Because this evaluates true. Would having something like (see the following code) do the trick?
```ts
labelClasses() {
  return {
    'visually-hidden': !!this.$attrs['aria-label']
  }
}
```

or

```ts
labelClasses() {
  return {
    'visually-hidden': this.$attrs.hasOwnProperty('aria-label')
  }
}
```

src/components/MaTextField/MaTextField.vue
alex-castelo: Nice approach 🔝

edufarre (approved):

PR #531, opened 2022-05-02
refactor: port ma-stack to ts

alex-castelo:


src/components/MaStack/MaStack.vue
```diff
@@ -68,17 +73,20 @@ export default {
...
+    } as PropOptions<StackElement>,
   },
 
-  render(createElement, { parent, props, slots, data }) {
-    const responsiveSpace = parent.$layout.getResponsivePropValue(props.space)
-    const responsiveAlign = parent.$layout.getResponsivePropValue(props.align)
+  render(createElement, { parent, props, slots, data }): VNode {
+    const responsiveSpace = parent.$layout.getResponsivePropValue(
+      props.space
+    ) as Spacing
```
edufarre, line 82: this `as Spacing` is really needed? Maybe what we should do instead if give a type to the return value of `getResponsivePropValue`

src/components/MaStack/MaStack.vue
```diff
@@ -11,12 +14,14 @@ const alignment = {
   right: 'flex-end',
 }
 
+type StackAlignment = keyof typeof alignment
```
edufarre, line 17: I'd rather have the values here, but it might be just a personal preference.

```suggestion
type StackAlignment = 'fill' | 'right'
```

src/components/MaStack/MaStack.vue
```diff
@@ -40,7 +45,7 @@ export default {
       type: [Array, String],
       required: true,
       validator: responsivePropValidator(Object.keys(spacing)),
-    },
+    } as PropOptions<Spacing>,
```
edufarre, line 48: I believe you missed that this prop can be an array as well
```suggestion
    } as PropOptions<Spacing | Spacing[]>,
```

src/components/MaStack/MaStack.vue
```diff
@@ -55,7 +60,7 @@ export default {
      * @values left, center, right, fill
      */
     align: {
-      type: [Array, String],
+      type: [Array, String] as PropType<StackAlignment>,
```
edufarre, line 63:
```suggestion
      type: [Array, String] as PropType<StackAlignment | StackAlignment[]>,
```

src/components/MaStack/MaStack.vue
alex-castelo: You're right! Thanks 👍🏽

src/components/MaStack/MaStack.vue
alex-castelo: done!

src/responsive.d.ts
```diff
@@ -5,7 +5,8 @@ declare type Breakpoint = 'mobile' | 'tablet' | 'desktop'
 
 type Layout = {
   readonly currentBreakpoint: Breakpoint
-  getResponsivePropValue: (values: string | string[]) => string
+  // eslint-disable-next-line no-unused-vars
+  getResponsivePropValue: <T>(values: T | T[]) => T
```
edufarre, line 9: 🚀

edufarre (changes requested):
I believe you should run the Vetur script again, as some properties still don't have the proper type. Once that is done, looking good to me 💪🏽

vetur/attributes.json
```diff
@@ -683,7 +674,7 @@
     ]
   },
   "ma-stack/space": {
-    "type": "array|string",
+    "type": "array",
```
edufarre, line 677: 🤔 Dunno why looks like we are losing one type (happens the same with other props)

alex-castelo: I think it's related with type inference, let me take a look

alex-castelo: The only reason I've found is the optional typing in the component --> `as PropType<Patata | Patata[]>` it seems to break `vetur script` .  I prefer to leave this this way due the final objective is to have TS support instead of Vetur, right?

What do you think about it?

edufarre (approved):

PR #532, opened 2022-05-02
refactor: port ma layout to ts

alex-castelo:
This is how I felt about this PR 😂 

image


src/components/MaLayout/MaLayout.vue
```diff
@@ -160,14 +209,23 @@ function renderGrid({ createElement, dom, grid, gap, justify, verticalAlign }) {
   return rows
 }
 
+interface RenderColumns {
```
sergigraciaj, line 212: Should be move this interface before the Vue.Extends, near the types, and rename it to IRenderColumns? Or just use it as a a type.

src/components/MaLayout/MaLayout.vue
```diff
@@ -160,14 +209,23 @@ function renderGrid({ createElement, dom, grid, gap, justify, verticalAlign }) {
...
+  createElement: CreateElement
+  columns: string[] | string
```
sergigraciaj, line 214: LayoutColumns type here? 🤔

src/components/MaLayout/MaLayout.vue
sergigraciaj: ^Same for RenderStack, RenderGrid

src/components/MaLayout/MaLayout.vue
```diff
@@ -108,14 +126,27 @@ export default {
...
-}
+})
 
-function getGrid({ parent, columns }) {
+function getGrid({
+  parent,
+  columns,
+}: {
+  parent: Vue
+  columns: string
```
sergigraciaj, line 136: Columns can be an Array too, so LayoutColumns type here and also on arguments in line 110, right?

src/components/MaLayout/MaLayout.vue
alex-castelo: We have talked before with Edu that Interface should be something that is extended or implemented. Let's keep the types for components for now (it's all debatable, I want to do some more research on this).

I agree to move everything up near the type declarations (even create a MaLayout type file).

sergigraciaj (approved):

PR #534, opened 2022-05-04
refactor: port ma datagrid to ts

alex-castelo:
## Context
I leave a couple of comments when reviewing this PR.

1) We said as an agreement that the types are defined in the component (in this case, as it is a component that uses different types, I have decided to create a `MaDatagrid.types.ts` file in the root of the component).

2) I have used any in some points where literally the value can be anything (I am open to suggestions).


src/components/MaDatagrid/MaDatagrid.vue
```diff
@@ -56,8 +60,12 @@ export default {
       type: Array,
       required: true,
       validator: (v) =>
-        v.every((column) => column.name && (column.value || column.component)),
-    },
+        v.every((column: Column) => {
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
edufarre, line 65: ?

src/components/MaDatagrid/MaDatagrid.types.ts
```diff
@@ -0,0 +1,16 @@
+import { VNode } from 'vue'
+
+interface IColumn {
```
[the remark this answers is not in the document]
alex-castelo: Matter of preference when I'm defining object shapes.

src/components/MaDatagrid/MaDatagrid.vue
alex-castelo: solved

src/components/MaDatagrid/MaDatagrid.types.ts
```diff
@@ -0,0 +1,16 @@
+import { VNode } from 'vue'
+
+type IColumn = {
```
sergigraciaj, line 3: If it's not an interface, let's rename the type so it don't start with an "I".

src/components/MaDatagrid/MaDatagrid.vue
```diff
@@ -61,9 +61,15 @@ export default Vue.extend({
...
-          column.name && (column.value || column.component)
+          /* Following this ESlint recomendations https://eslint.org/docs/rules/no-prototype-builtins */
+          if (Object.prototype.hasOwnProperty.call(column, 'value')) {
+            const valueColum = column as ValueColumn
+            return valueColum.name && valueColum.value
+          }
+          if (Object.prototype.hasOwnProperty.call(column, 'value')) {
+            const componentColum = column as ComponentColumn
+            return componentColum.name && componentColum.component
+          }
```
sergigraciaj, line 72: I'm lost here, why do we check the same twice? Can we remove the second check then? (not the code inside, just the if). The original validation does not work if we do `column.name && (column.value as ValueColumn || column.component as ComponentColumn)` ?

alex-castelo: I have tried it but it doesn't work. This type of if what it checks is that it is impossible for a ValueColumn to try to access the `component` property and vice versa.

PR #535, opened 2022-05-04
feat: Unify different ma-button and ma-link sizes

carlotamolintr:
We have MaButtons of different sizes in different apps. There are problems with the font size, the paddings and the horizontal styles. This PR fix the issue :)

I just added the width property to set full width in case we want the button to be in full width. By default the prop is set text-width.


src/components/MaLink/MaLink.vue
```diff
@@ -1,11 +1,14 @@
 <template>
-  <a :class="`ma-link--${category} ma-link`" :href="href" :rel="computedRel">
+  <a :class="[getClasses, 'ma-link']" :href="href" :rel="computedRel">
     <slot />
   </a>
 </template>
 
 <script lang="ts">
 import Vue, { PropOptions } from 'vue'
+import { Prop } from 'vue/types/options'
```
alex-castelo, line 9: I think that with `PropOptions`should be enough. This, is it being used?

src/components/MaButton/MaButton.vue
```diff
@@ -72,6 +72,15 @@ export default Vue.extend({
...
      */
+
+    /**
+     * Sets the button font-size style
+     */
+    size: {
+      type: String,
+      default: 'small',
+      validator: (val) => ['small', 'large'].includes(val),
+    },
```
alex-castelo, line 83:
```suggestion
    size: {
      type: String,
      default: 'small',
      validator: (val) => ['small', 'large'].includes(val),
    } as PropOptions<ButtonSize>,
```

In order to keep this TS inference, maybe we need to create this type on top of compoment. What do you think about it? :)

alex-castelo (approved):
I left a couple of comments! Good job Carlota 🚀

src/components/MaButton/MaButton.vue
carlotamolintr: Yes! I just forgot it... it didn't give me an error... weird...

src/components/MaLink/MaLink.vue
carlotamolintr: This was automatically imported I guess 🤔. I will remove it!

src/components/MaButton/MaButton.vue
alex-castelo: Maybe not an error. But if we want the IntelliSense features for autocomplete. I think we need it

sergigraciaj (approved):

xevops (approved):

src/components/MaButton/MaButton.vue
```diff
@@ -111,7 +111,7 @@ export default Vue.extend({
         'ma-button--rounded': this.rounded,
         [`ma-button--${this.category}`]: this.category,
         'ma-button--loading': this.loading,
-        [`ma-button--${this.size}`]: this.size,
+        [`ma-button--${this.size}`]: this.size && !this.rounded,
```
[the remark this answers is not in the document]
cillascillas: Same as [here](https://github.com/holaluz/margarita/pull/535#discussion_r896636563)

cillascillas (approved):

PR #536, opened 2022-05-05
refactor: Port MaHidden to typescript

edufarre:


src/components/MaHidden/MaHidden.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    return slots().default &&
+      shouldRenderContent(
+        props.above,
+        props.below,
+        parent.$layout.currentBreakpoint
+      )
+      ? h(component, slots().default)
+      : h()
+  },
```
alex-castelo, line 65: Checkmate Typescript ♟️ 🎩

src/components/MaHidden/MaHidden.ts
```diff
@@ -0,0 +1,99 @@
...
+      `neither 'below' nor 'above' prop were provided. Please provide at least one`
+    )
+  }
+}
+
+function shouldRenderContent(
+  above: AboveBreakpoints,
+  below: BelowBreakpoints,
+  currentBreakpoint: Breakpoint
+) {
```
alex-castelo, line 81: I don't know if we have talked about this but I believe that for Vue 2 and TS they recommend explicitly noting the return types of both methods and computed properties.

It is also true that surely, it already makes the correct type inference, so I think we would have to come to an agreement. 🤝

alex-castelo (approved):
🚀 🚀

src/components/MaHidden/MaHidden.ts
edufarre: This is not a method inside the component nor computed property, functional components does not have computed props

alex-castelo: Right! Sorry, my bad 😅

sergigraciaj (approved):

PR #542, opened 2022-05-11
Modify heading tokens

sergigraciaj:
[Figma branch](https://www.notion.so/holaluz/Update-headings-styles-70519fab37424f6f821a1872e5d02612)

Reordered the tokens in order to follow the order in the Figma file.
'letter-spacing' and 'text-transform' are just different in '2x-small' size, but I added as token to not have that logic in the component itself and if needed for future changes.


src/components/MaHeading/MaHeading.spec.ts
```diff
@@ -20,7 +20,9 @@ describe('MaHeading', () => {
     const { getByText } = renderComponent({ size: 'xsmall' })
     expect(getByText(slotText)).toHaveStyle({
       'font-size': '1.5rem',
-      'line-height': 1.35,
+      'line-height': '115%',
```
edufarre, line 23: I know it does work with percentages, but as far as I know we have always been setting `line-height` as `unitless values` do you think we could keep it like that?

edufarre (approved):

edufarre:
Shouldn't `text-transform: initial` be `text-transform: none`? 🤔  I don't really know

PR #555, opened 2022-05-25
Remove mabutton malink gradient

astyrdoncel:


edufarre (approved):

edufarre:
Beware that the target branch is `beta` and it should the rebranding one

PR #556, opened 2022-05-25
Holaluz new logo

astyrdoncel:
We can change the brand `brandImage` from the `manager.js` with the URL https://assets.holaluz.com/logos/logo-gradient.svg but we can't change the size of the logo. 

Result: 
Captura de Pantalla 2022-05-25 a las 10 49 46


sergigraciaj (commented):
Do we have this image in the assets repo? Would it make sense to get it from there so we have a single source of truth for svg's?

astyrdoncel:
> Do we have this image in the assets repo? Would it make sense to get it from there so we have a single source of truth for svg's?

Yes, we can use the assets image/svg. But the size of this svgs is huge and the storybook don't allow us resize this image. :S

PR #558, opened 2022-05-27
fix: Display proper hover and active color when using MaLink inside a MaAlert

sergigraciaj:
If we use a MaLink as a slot in the MaAlert (ideally a category no-background), as the MaLink specifies the hover and active color which is pink in its css, the link hover and active is displayed like that in the MaAlert, when it should be the same as when no state is applied. So in this PR we add the lines needed to fix that.

Current behavior for a MaLink inside a MaAlert

current-behaviour

Expected behaviour

expected-behaviour


carlotamolintr (approved):

src/components/MaAlert/MaAlert.spec.ts
```diff
@@ -58,4 +58,25 @@ describe('Alert', () => {
...
+      slots: {
+        default: `
+          <ma-link href="irrelevant">
+            This is a link
+          </ma-link>`,
+      },
+    })
+
+    const link = getByRole('link')
+    const mouseenter = new MouseEvent('mouseenter', {
```
edufarre, line 73: Would `fireEvent.mouseOver(...)` work here? (instead of `new MouseEvent`)

sergigraciaj: Yup, thx.

cillascillas (approved):

edufarre (approved):

PR #559, opened 2022-05-30
feat: Add number picker

cillascillas:
This PR adds [a brand new Number Picker defined in Watt](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=14906%3A9147)
Screenshot 2022-05-30 at 11 19 54

Assumptions made:
- Component adds and removes a defined step to/from initial value
- The value cannot be modified by keyboard/input (only by clicking in add/remove buttons)
- Add/Remove buttons are disabled if min/max are reached
- Component emits `change` event with new value when current value is changed
- Label is optional
- Suffix is mandatory and accepts singular/plural value
- Component logs error if the initial value is out of range
- Errors are handled via props as in the rest of form text inputs

This PR includes:
- New component and related tests
- New Remove (-) icon
- Vetur update

⚠️ **This PR includes hover in icons (inside buttons) handling because [this bug was detected](https://github.com/holaluz/margarita/issues/557) (once the bug is solved, this implementation can be deleted/refactored)**


src/components/MaNumberPicker/MaNumberPicker.vue
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Renders a number picker component following the Design System guidelines
+ *
+ * [Component's API documentation](https://holaluz.github.io/margarita/?path=/story/components-form-elements-number-picker--number-picker)
+ */
+export default Vue.extend({
+  name: 'MaNumberPicker',
+
+  components: {},
```
alex-castelo: I think this is not necessary 🕵🏽 .

src/components/MaNumberPicker/MaNumberPicker.vue
```diff
@@ -192,6 +195,40 @@ export default Vue.extend({
     isAddButtonDisabled(): boolean {
       return this.lazyValue === this.max
     },
+
+    // eslint-disable-next-line @typescript-eslint/ban-types
+    inputListeners(): Record<string, Function | Function[]> {
```
alex-castelo: Wow, just curiosity, why do we need to add these `inputListeners` I don't know if I fully understand 😃

alex-castelo (approved):
I've been leaving comments because I've been going through it commit by commit and in the end, they are all solved, I'm a mess 🤦🏽 .

I think it's super great! let's go with this NumberPicker 🔥

edufarre (approved):
Nice job! 🚀 

One question tho, when we implemented this on Negroni, we added a feature that handles the action when the user keeps pressing (maintaining) the button, and thus, the number change (see the video below).
Is this still a desired behaviour? cCuz we could add it (does not have to be in this PR) to keep consistency between projects, or we could stop using on Negroni otherwise.

https://user-images.githubusercontent.com/25011566/171337157-30dce0ab-864d-4f3d-93c3-ab790b35036e.mov

PR #560, opened 2022-05-31
fix: Display icons as proposed in Watt for MaAlerts

sergigraciaj:
Modify MaAlerts to display icon tones in their dark version (due to accesibility) as proposed in [Watt](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=2662%3A2403).

This PR is related to https://github.com/holaluz/assets/pull/42


src/components/MaAlert/MaAlert.spec.ts
```diff
@@ -53,9 +53,13 @@ describe('Alert', () => {
     expect(container.firstChild).toHaveClass('alert--red')
   })
 
-  test('icon has small size', () => {
+  test('icon has small size and proper color', () => {
```
edufarre, line 56: What does 'proper color' mean :trollface: 

What I mean by that, is that we could specify something like `Icon has small size and 'X' color by default`

src/components/MaAlert/MaAlert.vue
```diff
@@ -3,8 +3,8 @@
     <ma-icon
       class="icon"
       size="small"
-      :icon="icon"
-      :tone="tone"
+      :icon="icon[tone].icon"
+      :tone="icon[tone].tone"
```
edufarre, line 7: Could we have this in a computed? Dunno why, I do not like `icon[tone]` as syntactically would have to be something like `tones[tone].icon` and `tones[tone].color` (or `tones[tone].color`)
I'd rather have a better readability than complex templates

src/components/MaAlert/MaAlert.vue
```diff
@@ -61,18 +61,27 @@ export default Vue.extend({
...
-        case 'green':
-          return 'FeedbackSuccess'
-        case 'yellow':
-          return 'FeedbackWarning'
-      }
-      return 'FeedbackInfo'
-    },
+  data() {
+    return {
+      icon: {
```
edufarre, line 66: Could we change this? I believe `icon` is not the best naming (as commented in above). Maybe keeping the `computed` property way would work better, or perhaps just finding a better name for this would work too

edufarre (approved):

src/components/MaAlert/MaAlert.spec.ts
sergigraciaj: `dark color` then 😏

sergigraciaj:
Update: we decided with our design mates to modify the size of the icon. I also added a margin-top to align the icon a bit.

alex-castelo (approved):

PR #567, opened 2022-07-04
feat: close MaButton's API

sergigraciaj:
I been checking the MaButtons as I remember to have some issues while working with buttons, in fact everything seems to work fine, as long as we are not using a MaText inside the MaButton, if we do the state color of the text is not displayed properly. Alternatively we can use plain text or v-text using text that way works as a slot text too.
State's colors for MaIcon works perfectly thanks to the css [currentcolor](https://css-tricks.com/currentcolor/) property used in the MaIcon, which takes the color defined from the parent.

Anyways, though everything is working right now I propose a few changes to have an API closer to the proposed design in [Watt](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=5664%3A169)
The changes proposed are the addition of _icon_ and _text_ props (this is not a breaking change as we still accept a slot), but ideally we should work with props rather than slots in future developments.
The benefits from using these props are:
- Icon size changes automatically as defined in the MaButton size's prop.
- As API is closed, Icon will always be at the left with the same margin from the text.
- Is less verbose to define an icon and text (just write the props, no need to write components).


src/components/MaButton/MaButton.vue
```diff
@@ -5,15 +5,18 @@
     :class="[getClasses, 'ma-button']"
     @click="handleClick"
   >
+    <ma-icon v-if="icon" :icon="icon" :size="iconSize" :class="iconClass" />
+    {{ text }}
     <!-- @slot Button content slot -->
-    <slot v-if="!loading || !rounded" />
```
sergigraciaj, line 9: Does anyone know the reason of this peace of logic and if it's used somewhere?

alex-castelo: I understand it was so that the text would only appear if it wasn't loading or if it wasn't rounded. But I think it's fine if we remove it. 🤔

alex-castelo (approved):
👍🏽  Good job

src/components/MaButton/MaButton.vue
edufarre: Yeah, I would go ahead, and if needed, just add it again. But I don't think it is going to be the case. 🔥

edufarre (approved):

edufarre (commented):
BTW, this is a Breaking Change, isn't it? Because the MaButton's API  changes and now the button text has to be passed by using the `text` prop, right? Which rises a question in my op[inion, what would happen if we pass the `text` prop and use the `slot`? shouldn't one be overwriting the other? 🤔

edufarre:
> > BTW, this is a Breaking Change, isn't it? Because the MaButton's API changes and now the button text has to be passed by using the `text` prop, right? Which rises a question in my op[inion, what would happen if we pass the `text` prop and use the `slot`? shouldn't one be overwriting the other? 🤔
> 
> Nope, is not a breaking change as we still can use the slot. If we use the prop and the slot together both texts will appear inside the button, ideally the consumer should use just the prop one from now on.

Shall we add a message or a log when both of them are used at the same time? Or that maybe we recommend to use `text` instead the `slot`?

edufarre (approved):

PR #571, opened 2022-07-11
refactor: Port MaSelect to TS

alex-castelo:


src/components/MaSelect/MaSelect.vue
```diff
@@ -36,16 +36,29 @@
...
 </template>
 
-<script>
+<script lang="ts">
+import { defineComponent, PropType } from 'vue'
 import MaText from '@margarita/components/MaText'
 import uuid from '@margarita/utils/uuid'
 
+// eslint-disable-next-line @typescript-eslint/ban-types
+type ModelValue = string | number | boolean | Object
```
edufarre, line 45: Why `Object`? Not sure we should be using it here

src/components/MaSelect/MaSelect.vue
```diff
@@ -147,21 +160,23 @@ export default {
...
      */
     weight: {
-      type: String,
+      type: String as PropType<FontWeight>,
       default: '',
       validator: (value) =>
-        ['', 'bold', 'semibold', 'medium', 'regular'].includes(value),
+        ['', 'bold', 'semibold', 'medium', 'regular'].includes(
+          value as FontWeight
+        ),
```
edufarre, line 168:
```suggestion
 validator: (value: FontWeight) =>
        ['', 'bold', 'semibold', 'medium', 'regular'].includes(value),
```

src/components/MaSelect/MaSelect.vue
alex-castelo: I have the same doubt, what I thought was that as a `v-model` we have to be able to pass any kind of value of course this `Object` type only should be declared here and never as a generic type.

I'm open to suggestions!

src/components/MaSelect/MaSelect.vue
alex-castelo: Good catch!

src/components/MaSelect/MaSelect.vue
edufarre: I would not allow passing Object as value, don't even know if there is any use case through our project where this actually is used by passing an object as `v-model`. In order to do so, you needed to avoid TS, which is not a good practice.

sergigraciaj (approved):

PR #572, opened 2022-07-12
feat: MaModal with header image

astyrdoncel:
Modification of the MaModal component to be able to have an image inside the header:
- The size of the header changes to 200px always on desktop/tablet for the 'medium' and 'large' sizes
- In mobile we have a height of 150px
- In the case of the small modal it always has a height of 150px
- The close button has a color change outside the layout system when it is done the :hover event
- The text color becomes white (The background image will need to be adapted)
- The image should be created specifically for the size specified in Watt, if not, the entire width of the image is covered in the header
- [Fix] Adjusted the size of the button-close padding, which had been 'broken' with the update of the small/large button sizes

MaModal with image


src/components/MaModal/MaModal.css
```diff
@@ -82,3 +83,41 @@
     transition: transform 0.4s ease-in-out;
   }
 }
+
+.modal--with-image {
+  & .modal-header {
```
carlotamolintr, line 88: Could you please join the dot and the ampersand?

src/components/MaModal/MaModal.css
```diff
@@ -82,3 +83,41 @@
...
+    @media (max-width: 900px) {
+      height: 150px;
+    }
+  }
+
+  &.modal--width-small .modal-header {
+    height: 150px;
+  }
+
+  & .modal-title {
```
carlotamolintr, line 104: Could you please join the dot and the ampersand?

src/components/MaModal/MaModal.css
astyrdoncel: When you make a reference to a class that is inside another, not adjacent, you should not join the &.

src/components/MaModal/MaModal.css
```diff
@@ -48,6 +48,7 @@
 
 .modal-title {
   flex: 1;
+  padding-right: 1.5rem;
```
sergigraciaj, line 51: Not sure why this was not here before, but makes sense.
Anyways, could we use `var(--spacing-medium)` here?

src/components/MaModal/MaModal.spec.ts
```diff
@@ -27,6 +27,16 @@ describe('Modal', () => {
...
   })
 
+  test('accepts a image prop', async () => {
+    const { openModal, getByRole } = renderComponent({
+      props: { image: 'url' },
+    })
+
+    await openModal()
+
+    expect(getByRole('dialog')).toHaveClass('modal--with-image')
```
sergigraciaj, line 37: Would be cool if we assert the _imageURL_ and _headingTone_ here too

src/components/MaModal/MaModal.vue
```diff
@@ -18,19 +18,23 @@
...
+              data-testid="modal-header"
+            >
+              <ma-heading class="modal-title" size="xsmall" :tone="headingTone">
                 {{ title }}
               </ma-heading>
               <ma-button
                 ref="closeButton"
                 class="close-button"
-                size="small"
+                size="large"
```
sergigraciaj, line 37: Is this correct? I see _small_ defined in Watt

sergigraciaj (commented):
I'm missing this awesome feature in Storybook, could we add it, please?

src/components/MaModal/MaModal.vue
astyrdoncel: ✅ Fixed

src/components/MaModal/MaModal.css
astyrdoncel: ✅ Fixed

src/components/MaModal/MaModal.css
```diff
@@ -82,3 +83,44 @@
...
+.modal--with-image {
+  & .modal-header {
+    position: relative;
+    min-height: 200px;
+    height: 200px;
+    background-size: cover;
+    background-repeat: no-repeat;
+    padding-bottom: var(--spacing-medium);
+
+    @media (max-width: 900px) {
```
xevops, line 96: It is possible to use breakpoint tokens sizes defined within margarita?

Moreover, I would suggest use min-width instead of max-width when using media-queries to follow a mobile first approach

astyrdoncel: We have two breakpoints:
- --breakpoint-tablet: 900;
- --breakpoint-desktop: 1440;

But I don't know who we can use for this kind of case, because var(--breakpoint-tablet) cannot be concatenated with `+'px'`.

Should we change the current breakpoints? (or add new)

sergigraciaj (approved):
🔝

carlotamolintr (approved):

xevops (approved):

PR #582, opened 2022-08-10
refactor: Port MaModal to typescript

edufarre:
Beware that I added the `Volar` extension as recommended (along with `prettier` and `eslint`) because `Vetur` does **NOT** support `Vue v2.7`.


alex-castelo (approved):
I don't have proof but neither have doubts 🧙🏽 🐉 🎈 Great work

sergigraciaj (approved):

PR #589, opened 2022-08-31
feat: add required prop to form elements

cillascillas:
# Description
This PR adds a new prop: **required** to form elements: MaTextField (and all the variants), MaOption and MaSelect.

**Required** can be `true` or `false` but it is `false` by default
- False: Nothing changes
- True: an * is added to the label

Screenshot 2022-08-31 at 14 32 37

## Assumptions made in this approach
- Margarita does not contain logic of the components: For this case that means that required prop only adds the * to the label but does not make any behaviour change. The responsibility on how to manage it is from the form that would include the field. (KaForm is going to experiment some changes due to this new addition)

## Doubts/Decisions to be made by reviewers
- Following this approach screen readers would announce this field as: _"Full Name Asterisk"_ which an experienced user will probably know what this is intended for, but many users will have no clue. The solution for this could be adding an `aria-hidden` with an explanation, however since we do not manage translations of texts in margarita, it was not added. (Maybe adding a text for required prop? Would it be too much?)
Apparently, and regarding to the design proposal a legend in the bottom of the form should be explaining what the asterisk means, and it should cover this problem. But as we know, that would mean that when defining the form, the developer should include it manually.
👉 Is there any better approach you think of? Proposals are welcome!!!

- MaOption (Radio): Do we really need the required prop? When do we need to remind that it is required since it is usually placed in group and not all options should be marked with the *? Maybe is the question or label wrapping all options the one that should have it?
👉 Your thoughts?
 
### References
Docs or pages used to make this decisions:
- [Accesibility developer guide](https://www.accessibility-developer-guide.com/examples/forms/required/#using-html-5-client-side-validations)
- [A11y-101](https://a11y-101.com/development/required)


alex-castelo (approved):
> Following this approach screen readers would announce this field as: "Full Name Asterisk" which an experienced user will probably know what this is intended for, but many users will have no clue. The solution for this could be adding an aria-hidden with an explanation, however since we do not manage translations of texts in margarita, it was not added. (Maybe adding a text for required prop? Would it be too much?)
> Apparently, and regarding to the design proposal a legend in the bottom of the form should be explaining what the asterisk means, and it should cover this problem. But as we know, that would mean that when defining the form, the developer should include it manually.
> 👉 Is there any better approach you think of? Proposals are welcome!!!

IMHO, if the design proposal has the legend solution, the implementation has to be as similar as possible and maybe, for now, we can test if the component is really used as it should be in each context or if maybe the idea of including the description in `alria-hidden` is better.

> MaOption (Radio): Do we really need the required prop? When do we need to remind that it is required since it is usually placed in group and not all options should be marked with the *? Maybe is the question or label wrapping all options the one that should have it?
> 👉 Your thoughts?

I totally agree with you. In the case of Ma-Option, I believe that the required element should be the label, not the option itself.

Thank you for your very clear explanation and for the questions raised.

julioacontreras (approved):

icurdu:
> Following this approach screen readers would announce this field as: "Full Name Asterisk" which an experienced user will probably know what this is intended for, but many users will have no clue. The solution for this could be adding an aria-hidden with an explanation, however since we do not manage translations of texts in margarita, it was not added. (Maybe adding a text for required prop? Would it be too much?)
Apparently, and regarding the design proposal a legend in the bottom of the form should be explaining what the asterisk means, and it should cover this problem. But as we know, that would mean that when defining the form, the developer should include it manually.

IMHO, I think we should add the aria-hidden text in Spanish, meanwhile we don't have translation in Margarita (I know this is not ideal, but it covers a high percentage of users). Maybe having a default value replaceable with the translations JSONS in every project would be enough too. What I'm sure is having the legend is not enough to tell the user with disabilities that some labels are required. 


> MaOption (Radio): Do we really need the required prop? When do we need to remind that it is required since it is usually placed in group and not all options should be marked with the *? Maybe is the question or label wrapping all options the one that should have it?

The required prop in Ma-Option Radio should be placed in the label since it's a component that is always required, but it's managed by the group, it's never going to go by itself. 
You can see in [Figma](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=17923%3A10088) that we defined it this way :)

Thanks for spotting and sharing these doubts!

astyrdoncel:
These are the final changes made for this new feature:

>Following this approach screen readers would announce this field as: "Full Name Asterisk" which an experienced user will probably know what this is intended for, but many users will have no clue. The solution for this could be adding an aria-hidden with an explanation, however since we do not manage translations of texts in margarita, it was not added. (Maybe adding a text for required prop? Would it be too much?)
Apparently, and regarding to the design proposal a legend in the bottom of the form should be explaining what the asterisk means, and it should cover this problem. But as we know, that would mean that when defining the form, the developer should include it manually.
👉 Is there any better approach you think of? Proposals are welcome!!!

**The title "Campo requerido" is added in order to help people with disabilities. In the future, it may be considered to be able to pass translations to the component/Margarita.**

> MaOption (Radio): Do we really need the required prop? When do we need to remind that it is required since it is usually placed in group and not all options should be marked with the *? Maybe is the question or label wrapping all options the one that should have it?
👉 Your thoughts?

**The possibility of being able to put the required field (*) is eliminated when it is a MaOption of type radio.**

PR #591, opened 2022-09-14
refactor: Port MaTabs component to TS

edufarre:


src/components/MaTab/MaTab.vue
```diff
@@ -1,35 +1,79 @@
...
+      from: 'selectedTabName',
+      default: '',
+    },
+    hoveredTabName: {
+      from: 'hoveredTabName',
+      default: '',
+    },
+    tabParentId: {
+      from: 'tabParentId',
+      default: '',
```
alex-castelo, line 43: Nice point. First time I see `provide / inject` pattern in use 🔝

src/components/MaTab/MaTab.vue
```diff
@@ -1,35 +1,79 @@
...
+import { defineComponent, PropType } from 'vue'
+import MaTabButton from './_components/MaTabButton'
+import type { Icon } from '../MaIcon/availableIcons'
+
+export type MaTabProps = {
+  title: string
+  name: string
+  icon: Icon | null
+  pill: string | number | null
+}
```
alex-castelo, line 25: 🎩

alex-castelo (approved):
Congrats 👏🏽  It is a difficult PR to follow because of the difficulty involved, but still a very clear and good job.

I tried it 2 or 3 times and ended up a little disoriented so congratulations. I leave here some interesting links in case people who check the PR are curious to see some less common parts.

- [Link to provide / inject pattern of vue](https://vuejs.org/guide/components/provide-inject.html) (this is explained with Vue 3 but the idea is the same for Vue 2)

pd: the tests pass, I have manually checked that it works, so, here we go! 🍾

src/components/MaTab/MaTab.vue
alex-castelo: Me sombrero el quito

aescri (approved):

PR #597, opened 2022-10-05
Feat: Add file loader component

IvanSanNicolasVolcanic:


alex-castelo:
We need to run vetur script

alex-castelo (approved):
Awesome 🚀

edufarre (approved):

icurdu:
Massive work, mates. Thanks for make this component alive! 
There are some enhancements we can tackle in other releases, feel free to create an issue if you want.

- Accessibility: does not have focus
- Drag and drop state
- Animations between states

ferbg:
Good job 🚀 
Just a question, what is the prop `label` for? Is it to be used as `tag` in MaText? If so, I think it needs to be assigned correctly (`:tag="label"`)

edufarre:
> Good job 🚀 Just a question, what is the prop `label` for? Is it to be used as `tag` in MaText? If so, I think it needs to be assigned correctly (`:tag="label"`)

I believe it already has it: https://github.com/holaluz/margarita/pull/597/files#diff-6230db1e1551538d07f6d9c65739086682125e15a7377eaad8947aaada49b060R4

>Massive work, mates. Thanks for making this component alive!
There are some enhancements we can tackle in other releases, feel free to create an issue if you want.

We did not implement any of it, because the component on your side still needed review. This is the first approach to the file input, so as you said, we can always iterate it in the future. However, I would sort out the `focus`, as it is part of the WAI

alex-castelo:
> Accessibility: does not have focus

Solved, @icurdu 🕺🏽

PR #598, opened 2022-10-10
feat: 2x-small added

ferbg:
2x-small added
image


edufarre (commented):
I believe you should run `tokens:build` in order to update the related token styles

edufarre (approved):

sergigraciaj (approved):

PR #602, opened 2022-10-13
fix: Avoid duplicating the event when modal window closes

ferbg:
MaModal has an issue, when the modal is closed, the event 'close' is emitted twice:
* Calling `closeModal` function by ['close' button](https://github.com/holaluz/margarita/blob/v16.3.0/src/components/MaModal/MaModal.vue#L45) / [ESC key](https://github.com/holaluz/margarita/blob/v16.3.0/src/components/MaModal/MaModal.vue#L184) / [mousedown hook](mousedown.self=)
* Calling `closeModal` function by [after-leave hook](https://github.com/holaluz/margarita/blob/v16.3.0/src/components/MaModal/MaModal.vue#L11)


src/components/MaModal/MaModal.stories.js
```diff
@@ -32,7 +36,10 @@ const Template = (args, { argTypes }) => ({
         <ma-stack space="medium" align="left">
           <ma-text size="small">My amazing content</ma-text>
           <ma-text-field label="And a random input" type="text" />
-          <ma-button @click="closeModal">close</ma-button>
+          <div style="display: flex; gap: var(--spacing-small); justify-content: flex-end;">
```
astyrdoncel, line 39: ⚠ css inline? Can you use a ma-layout or ma-stack to separate elements? or as a last option use a class :)

astyrdoncel: ok I thought it was from the component itself. I don't know if it's common in stories to use css inline hehe

ferbg: in fact, I used inline styling to align the buttons to the end and forgot a `width:100%` 😅​🤭​

alex-castelo (approved):

astyrdoncel (approved):

PR #603, opened 2022-10-17
fix: title spacing in ma-modal

carlotamolintr:
This pr is to fix the title in ma-modal image component. As you can see in the image, the title letters are cutted in some cases (please see "tejado"). So, I just added a small padding to fix it.

Captura de pantalla 2022-10-17 a las 16 57 10

You can check through controls the result.
I leave a link image here to try --> https://pr-276.d14ufzzq81ygsn.amplifyapp.com/_nuxt/img/holaluz-woman.bf97da9.jpg


src/components/MaModal/MaModal.css
```diff
@@ -106,6 +106,7 @@
 
   & .modal-title {
     padding-right: 0;
+    padding-bottom: 0.25rem;
```
edufarre, line 109: dunno if this is the best approach here. We bumped into a similar issue a few weeks ago and it was because of the `crop` attributes of `ma-text`/`ma-heading`. Maybe it is worth looking a bit more in-depth and seeing if we can fix it for all the cases, not just for the modal title.

carlotamolintr: Do you mean to change crop value? I would like to know which other cases are affected. I just checked previous releases and I couldn't find it. If anyone knows, please tell me :)

edufarre (approved):

astyrdoncel (approved):

PR #607, opened 2022-10-25
fix: green base new value

ChaoHu99:
New value for accessibility green base token


edufarre (commented):
I belive you forgot to trigger the `tokens:build` command

edufarre (approved):

carlotamolintr (approved):

PR #611, opened 2022-11-09
feat: Add footer to MaModal

ferbg:
New slot for the modal footer
image


icurdu (approved):
Everything looks good to me! In the future, we may need to do a small variation of the footer template when the modal is size="small", because the texts of the buttons can be larger than the container and make buttons in 2 lines (I just realized of that behavior). The proposed solution is simple, just stacking the buttons and make it full-width, but this requires a complex architecture using ma-layout. 
Captura de pantalla 2022-11-09 a las 10 33 51
Captura de pantalla 2022-11-09 a las 10 43 35

edufarre (commented):
Could we add some tests asserting that the footer is rendered and has the expected behaviour?

edufarre:
>The proposed solution is simple, just stacking the buttons and make it full-width, but this requires a complex architecture using ma-layout.

If we already know that the expected styling of the modal's footer buttons should be stacked instead of in a row, I would tackle it now, even though we do not use `ma-layout` and style them using CSS

astyrdoncel (approved):

PR #614, opened 2022-11-11
fix: Fix MaTabs dynamic content rendering

### LINKED ISSUE #610: fix: Ma Tabs behaviour (not rendering nested components)
It seems that MaTabs only render direct nodes not nested ones. I don't know if this is the expected behaviour.

I noticed this when I was triying to update Margarita version here:
https://github.com/holaluz/bonasera/pull/1845

edufarre:
Closes #610


alex-castelo (approved):
Super! I love it. I have added some attributes following the `ARIA` standards to have a more accessible component.

I just have a doubt about the `tabpanel` section. I don`t know if it`s a good practice to destroy and regenerate each of them or it's better to have several `tabpanels` divs (as many as tabs) so that users know their existence.

I think this discussion is something that could be handled as a future enhancement. I have left some useful links:

- [How to create a tab component with Vue](https://learnvue.co/tutorials/vue-reusable-tabs-component)
- [Tab role MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role)

src/components/MaTabs/MaTabs.vue
```diff
@@ -3,23 +3,33 @@
...
       <ma-tab-button
         v-for="tab in tabs"
+        :id="tab.name"
         :key="tab.name"
+        :aria-controls="`${tab.name}`"
         :name="tab.name"
         :title="tab.title"
         :icon="tab.icon"
         :pill="tab.pill"
+        role="tab"
```
edufarre, line 13: Maybe we could move this into the `ma-tab-button` component

astyrdoncel (approved):

PR #626, opened 2022-11-29
feat: Create ma skeleton component

carlotamolintr:
_UPDATED!!!_

Hello all,

At least we created two components:

**MaSkeleton:**:

This component have the following props:
- `type`: By default the value is `button`. The Possible values are: `heading`, `text`, `picto`,  and `button`.
- `size`: By default the value is 'medium'. The prop sets the size according to the component type that we want replace.
-  `width`:  Can be used as percentage or pixels. By default is 100%.

**MaSkeletonCustom**:
- `border`: Default as `''`. 
- `height`: This is a required prop. Can be used as percentage or pixels. 
-  `width`:  This is a required prop.  Can be used as percentage or pixels. 

In the documentation section. You can read how the component works. 

Comments are welcome 🤗 

[Figma documentation](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/branch/eobFLxR3LKfTWtVAaDKWmG/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=21508%3A11733&t=OoXfCHjb8yO8sCu8-0)


icurdu:
General feedback:
- Beautiful documentation! 
- The transition is too fast compared to the Figma prototype. In the design file the transition is setted 1s between every state (with 0.1s of delay), so the animation-duration in code should be 2s (or 2.2s if you want to achieve 100% accuracy).

About the implementation (I'm not a dev so take this comments 
- I think it will be hard for the dev to remember which are the allowed values for each 'type', because not all the values can apply to one another. Maybe there's another way to tacking this and avoiding this prop dependency. 

Thanks for taking time to tackle this feature!

carlotamolintr:
> General feedback:
> 
> * Beautiful documentation!
> * The transition is too fast compared to the Figma prototype. In the design file the transition is setted 1s between every state (with 0.1s of delay), so the animation-duration in code should be 2s (or 2.2s if you want to achieve 100% accuracy).
> 
> About the implementation (I'm not a dev so take this comments
> 
> * I think it will be hard for the dev to remember which are the allowed values for each 'type', because not all the values can apply to one another. Maybe there's another way to tacking this and avoiding this prop dependency.
> 
> Thanks for taking time to tackle this feature!

@icurdu Thank you Iban for your comment. I agree with you that is a bit hard to remind all the specifications of the component. Maybe we can split in more components as `ma-skeleton`, `ma-skeleton-typo`, `ma-skeleton`picto, `ma-skeleton-button`. About this, let see what the other FE think :)

Regarding the transition. I thought it was the one that is in Watt😅. Sorry, I will check it and replaced for the good one.

alex-castelo:
First of all! Congrats, such a great job! 🎉 

I like your approach and have some comments about your questions.

> QUESTION: Is possible to set required prop depending on another prop? Since I have investigated, the response is not... But just in case...

I think it is not possible but, if you want to set a value depending on another you can use a `computed` 💡 

IMHO, I think there are infinite approaches to handle this, but two comes to my main:

The first one, is to create a `skeleton` component totally flexible with props:
- width
- height
- border-radius

This way when you use the component you have full flexibility to customize it. (Also if you use `%` it is responsive)

The second one, is to create custom components for each `type` of skeleton, maybe with the props you've used in this PR.
Maybe this option will speed up the development and will work for the majority of cases.

The third one is to do both. 😂 

Anyway, good job! Let's discuss this in the FE Crew!

carlotamolintr:
Thanks for you comment @alex-castelo! 


Related with:
> > QUESTION: Is possible to set required prop depending on another prop? Since I have investigated, the response is not... But just in case...
> 
> I think it is not possible but, if you want to set a value depending on another you can use a `computed` 💡
> 

Yes, I will do it like that if we finally want this "super" component manage all casuistry. 

Regarding the different approaches of the component... For me will be better to have both. Let's see what other people says. :)

src/components/MaSkeleton/MaSkeleton.vue
```diff
@@ -0,0 +1,131 @@
...
+
+    border: {
+      type: String,
+      default: '',
+      validator: (border: Border) =>
+        ['', 'small', 'medium', 'large'].includes(border),
+    } as PropOptions<Border>,
+  },
+
+  render(createElement, { parent, props, data }): VNode {
```
edufarre, line 65: Crazy question, why did you use the render function here instead of the template? There is this big function with some `if`s , which they might be easier to understand using computed properties, or am I missing something?

carlotamolintr:
> Overall, it looks promising, I left some comments here and there.
> 
> However, there is one thing I do not like: the component will have a different API (usage of its properties). Let me explain: I believe it is quite odd to have a property `width` that will be `required` in a use case, but not in other ones... Could we please try to find out a common use case and go from there? I say so because this is making me feel like we are trying to wrap as much as we can inside this component and it could lead to misleading usage or behaviour.
> 
> PS: Love the docs btw 😄

Hi @edufarre, Thank you for your comments. 

Seeing all the comments above, I will split the component in two. So, I hope will be easier to use it. At least, we will have two skeletons:

1. First component will be more customisable and will accept `width`, `height` and `border` props. Do you have any proposals for namings? I propose  -->`ma-skeleton-custom`.
2. Second component will need a required `type`.  An specific `size` could be defined (by default: 'medium'). The size prop will manage the height. And also, it will have a`width`by default setted as 100%. Which name can we use here?

Any thoughts or comments will be welcome 😬

src/components/MaSkeleton/MaSkeleton.vue
```diff
@@ -0,0 +1,131 @@
...
+        class: {
+          [`ma-skeleton--picto-${props.size}`]: props.size,
+          ['ma-skeleton--picto']: props.type,
+        },
+      }
+      return createElement('div', mergeData(data, pictoData))
+    }
+
+    if (props.type === 'button') {
+      // TODO: QUESTION: What do you thing about setting by default: size = small ?? In the other types, the default size is medium.
```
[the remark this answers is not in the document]
icurdu: size="small" should be the default size, because is the default size of Ma-Button (is the most common prop among all projects)

src/components/MaSkeleton/MaSkeleton.vue
```diff
@@ -0,0 +1,131 @@
...
+        style: {
+          width: props.width !== '' ? props.width : 'auto',
+        },
+      }
+      return createElement('div', mergeData(data, pictoData))
+    }
+
+    const size = parent.$layout.getResponsivePropValue(props.size) as Size
+
+    const skeletonTypo =
```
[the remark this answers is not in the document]
icurdu: SkeletonText (for text and headings) could be a better name, but in Figma we have SkeletonTypography (I know, it's too large, we can change it in Watt in favor of the one you choose :)

sergigraciaj:
IMO we need to simplify:
The Skeleton should only be seen for seconds, ideally our systems should be efficient enough to minimize the use of Skeletons, also we have another approach for handling these situations using loaders/spinners.
Taking this into account, I think we can simplify the Skeleton from the design to the implementation, we don't need to have skeletons exacts to what we are going to paint.
First of all we can handle the width with MaColumns/MaLayout being full-width the default, then checking the proposal in Figma, I see four concepts: heading/text, buttons, icons/pictos and blocks (alerts, cards, images...).
So we can have this in four components with just a height prop (small, medium for button and icon/picto or small, medium and large for text and block). Having these four components also removes the usage of the border prop as it is defined in the component itself.

src/components/MaSkeleton/MaSkeleton.vue
```diff
@@ -0,0 +1,109 @@
...
+  name: 'MaSkeleton',
+
+  props: {
+    /**
+     * Sets the element type you want to reproduce
+     */
+    type: {
+      type: String,
+      default: 'button',
+      // required better?
```
carlotamolintr, line 35: Do you agree to set the button by default? Or is better to put required?

sergigraciaj: IMO required, so we know explictly the type when composing skeletons

src/components/MaSkeletonCustom/MaSkeletonCustom.vue
```diff
@@ -0,0 +1,57 @@
...
+     *  Sets the height in px. Required for empty type. For the other types, height is automatically setted (depending on size)
+     */
+    height: {
+      type: String,
+      required: true,
+    },
+
+    /**
+     *  Sets the border-radius just valid for the empty type.
+     */
```
sergigraciaj, line 39: All the props decriptions needs to be updated as there is no empty type, right?

src/components/MaSkeletonCustom/MaSkeletonCustom.stories.js
```diff
@@ -0,0 +1,42 @@
...
+    },
+    height: {
+      control: 'text',
+      defaultValue: '100px',
+    },
+  },
+}
+
+const Template = (args, { argTypes }) => ({
+  components: { MaSkeletonCustom },
```
sergigraciaj, line 31: we need this property? :o

src/components/MaSkeletonCustom/MaSkeletonCustom.spec.js
```diff
@@ -0,0 +1,35 @@
...
+    })
+
+    expect(getByTestId('skeleton-custom')).toHaveClass('ma-skeleton-custom')
+    expect(getByTestId('skeleton-custom')).toHaveStyle({
+      width: '100px',
+      height: '100px',
+    })
+  })
+
+  test(`renders the correct border and height when setted as prop`, () => {
```
sergigraciaj, line 18: not checking the height here, just the border

src/components/MaSkeleton/MaSkeleton.stories.js
```diff
@@ -0,0 +1,42 @@
...
+    },
+    width: {
+      control: 'text',
+      defaultValue: '100%',
+    },
+  },
+}
+
+const Template = (args, { argTypes }) => ({
+  components: { MaSkeleton },
```
sergigraciaj, line 33: we need this property? :x

src/components/MaSkeleton/MaSkeleton.vue
carlotamolintr: I will change it. :)

src/components/MaSkeletonCustom/MaSkeletonCustom.spec.js
carlotamolintr: I will remove it. Thanks :)

src/components/MaSkeleton/MaSkeleton.stories.js
carlotamolintr: @sergigraciaj wihich one?

sergigraciaj (approved):

Veenders (approved):

docs/components/MaSkeleton.docs.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### Picto type 
+
+Picto type needs `size` value. Valid values are: `xsmall`, `small`, `medium`, `large` and `xlarge`
+Here you can see an example:
+
+<Story id="components-skeleton-examples--picto-type" />
+
+```js
+// Code:
```
astyrdoncel, line 33: I think it's redundant

docs/components/MaSkeleton.docs.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### Button type 
+
+Button type needs `size` prop. And the valid values are `small` or `large`. Also the `width` could be customized. If not set, will be full-width skeleton-button.
+Here you can see an example:
+
+<Story id="components-skeleton-examples--button-type" />
+
+```js
+// Code:
```
astyrdoncel, line 53: I think it's redundant

docs/components/MaSkeleton.docs.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### Text type 
+
+Text type needs `size` prop. And the token will set the proper height automatically. The valid values for size are: `xsmall`, `small` or `medium`. Also the `width` could be customized. If not set, will be full-width skeleton-text. The width uses percentatge (the ´%´ is no needed).
+Here you can see an example:
+
+<Story id="components-skeleton-examples--text-type" />
+
+```js
+// Code:
```
astyrdoncel, line 74: I think it's redundant

docs/components/MaSkeleton.docs.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+### Heading type 
+
+Heading type needs `size` prop. And the token will set the proper height automatically. The valid values for size are: `2x-small`, `xsmall`, `small`, `medium`, `large`, `xlarge`. Also the `width` could be customized. If not set, will be full-width skeleton-text. The width uses percentatge (the ´%´ is no needed).
+Here you can see an example:
+
+<Story id="components-skeleton-examples--heading-type" />
+
+```js
+// Code:
```
astyrdoncel, line 91: I think it's redundant

docs/components/MaSkeletonCustom.docs.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+
+We will see more deeply on the nexts steps.
+
+Here you can see an example:
+
+<Story id="components-skeletoncustom--skeleton-custom" />
+
+```js
+// Code:
```
astyrdoncel, line 21: I think it's redundant

src/components/MaSkeleton/examples/examples.stories.js
```diff
@@ -0,0 +1,89 @@
...
+    </ma-layout>
+  </div>`,
+})
+
+export const ButtonType = () => ({
+  title: 'Button Skeleton',
+  template: `
+  <div>
+    <ma-layout gap="medium" columns="6 6" vertical-align="center" style="width: 600px">
+      <h3>Size small</h3>
```
astyrdoncel, line 26: **Or is "Size: small" or "Small size".** What do you think?

In this case you have put the name of the size together with the skeleton, I suppose that to be able to compare them. I think that the same thing could be done with the rest of the skeleton variants and do something like:

Captura de Pantalla 2023-01-31 a las 10 42 10

src/components/MaSkeleton/MaSkeleton.vue
```diff
@@ -0,0 +1,108 @@
+<template>
+  <div
+    data-testid="skeleton"
+    class="ma-skeleton"
+    :class="getClasses"
+    :style="getStyles"
+  ></div>
```
astyrdoncel, line 7:
```suggestion
  />
```

astyrdoncel:
Hello!! First of all good job Carlota! 😄 

📖  Missing from the documentation to see the **figma documentation** of the component (to see the best practices to be expected). You have an example of how we did it in the Stepper [here](https://601199d127273f0021b037b0-ixrlymvsiu.chromatic.com/?path=/docs/components-stepper--stepper). 🔗 [Skeleton Figma](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=21470%3A8818&t=GmtYcS7funOfBeRw-0)

💬  I have left a couple more comments to improve the documentation.

docs/components/MaSkeleton.docs.mdx
carlotamolintr: I like it!

docs/components/MaSkeleton.docs.mdx
carlotamolintr: I like it!

docs/components/MaSkeleton.docs.mdx
carlotamolintr: I like it!

docs/components/MaSkeleton.docs.mdx
carlotamolintr: I like it!

docs/components/MaSkeletonCustom.docs.mdx
carlotamolintr: I like it!

src/components/MaSkeleton/MaSkeleton.vue
carlotamolintr: I just fix it!

PR #627, opened 2022-11-29
BREAKING CHANGE: Update ma stepper [POC]

TamaraRodriguez:
Hi all,

Here is a POC for the new Stepper component. The goal of the steppers is to show progress through a sequence of logical, numbered steps vertically and horizontally.

The `MaStepperPOC.vue` component is the main component where the general styles are defined, passing as props: `mode` (horizontal/vertical), `tone` (light/dark) and `selected` (step indicator). The content must be passed to this component by `slot`. In order for the content to be displayed correctly `MaStepperStep` must be used passing the information inside it.

The `MaStepperStep.vue` component is the step where we define the following props: `picto`, `index`(step number), `title`, `description` and `info`. Depending on the needs of use, we could add any content (form, texts, lists, etc.) by means of a slot, substituting-nullifying the mentioned props.

The proposed development idea is to reuse the same component for horizontal (responsive version) and vertical formats.

**QUESTION**: Should the components be independent according to the format?

In the examples stories you can check the caustics contemplated according to design.

You can learn more in the [figma documentation](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=22788%3A9857&t=PGp9okqiNOHxxQSc-1).

Any opinion will be welcome, everything adds up.

Captura de pantalla 2022-11-29 a las 17 08 32


src/components/MaStepperStep/MaStepperStep.vue
```diff
@@ -0,0 +1,230 @@
...
+      return this.tone === 'light' ? 'gray-dark' : 'white'
+    },
+    pictoSize() {
+      return this.$layout.currentBreakpoint !== 'mobile' ? 'small' : 'xsmall'
+    },
+    circleClass() {
+      const isActive = this.isSelected ? 'ma-stepper-step__circle--active' : ''
+      const isDark = this.tone === 'dark' ? 'ma-stepper-step__circle--dark' : ''
+      return ['ma-stepper-step__circle', isActive, isDark]
+    },
```
edufarre, line 119: We could use the same approach that is suggested into another comment, instead of giving `''` as a class, using the Object format for the return type

edufarre (commented):
`src/components/MaStepperPOC/MaStepper.css` is an empty `css` file. I believe we should do one of the following, delete the file or move its CSS in it.

⚠️ Bear in mind that some files and classes have `POC` in its name, and we should remove it.

carlotamolintr:
Hi all :),

It looks good! By the way, I would like to propose you to change two props names:

- Prop `mode` instead of `tone`. As the values just could be light/dark it reminds me to the dark/light mode of screens.
- Prop `position` instead of `mode`. Cause we can have to positions values 'vertical' or 'horizontal'

src/components/MaStepperPOC/MaStepperPOC.vue
```diff
@@ -0,0 +1,85 @@
...
+        this.mode === 'vertical' || this.$layout.currentBreakpoint === 'mobile'
+          ? 'ma-stepper-poc--vertical'
+          : ''
+      return ['ma-stepper-poc', verticalClass]
+    },
+  },
+}
+</script>
+
+<style scoped>
```
carlotamolintr, line 60: You can move this css to css file (currently empty)

src/components/MaStepperPOC/MaStepperPOC.vue
```diff
@@ -0,0 +1,85 @@
...
+.ma-stepper-poc {
+  position: relative;
+}
+
+.ma-stepper-poc:not(.ma-stepper-poc--vertical)
+  > div
+  >>> .ma-stepper-step:last-child::before {
+  content: none;
+  background: pink;
```
carlotamolintr, line 69: We need to use the color var here :)

src/components/MaStepperStep/MaStepperStep.vue
```diff
@@ -0,0 +1,230 @@
...
+  name: 'MaStepperStep',
+  inject: ['$mode', '$selected', '$tone'],
+  props: {
+    index: {
+      type: Number,
+      required: true,
+    },
+    picto: {
+      type: String,
+      default: undefined,
```
carlotamolintr, line 65: I always use `default: ''` when is string type. But let's see what other people says 😬

sergigraciaj (commented):
The POC overall looks good, I would rename the files and remove the skips from the tests to check the PR properly.
Also we are missing the size prop, and the tone of the info text is no the defined in the design.

src/components/MaStepperStep/MaStepperStep.vue
```diff
@@ -0,0 +1,230 @@
...
+        return this.isColored ? 'white' : 'gray'
+      return this.isColored ? 'pink' : 'gray'
+    },
+    pictoTone() {
+      if (
+        this.tone === 'dark' &&
+        (this.mode === 'vertical' ||
+          (this.mode === 'horizontal' &&
+            this.$layout.currentBreakpoint === 'mobile'))
+      )
```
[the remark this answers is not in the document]
ferbg: done (or so I think). I have simplified the `mode` property in this component, the container will set the mode taking into account the breakpoints

src/components/MaStepperPOC/MaStepperPOC.vue
```diff
@@ -0,0 +1,85 @@
...
+</script>
+
+<style scoped>
+.ma-stepper-poc {
+  position: relative;
+}
+
+.ma-stepper-poc:not(.ma-stepper-poc--vertical)
+  > div
+  >>> .ma-stepper-step:last-child::before {
```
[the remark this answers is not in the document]
ferbg: This is the way we found to control the end line from the container:
image
image
any better ideas?

PR #628, opened 2022-11-30
fix: update ma-link secondary with white-base background color

astyrdoncel:


ferbg (approved):

carlotamolintr:
If the link background was already defined in Watt (is nothing new), will be a fix instead of a feat. :)

carlotamolintr (approved):

PR #632, opened 2022-12-23
fix: MaModal footer padding when there is no content

astyrdoncel:


src/components/MaModal/MaModal.vue
```diff
@@ -148,6 +148,9 @@ export default defineComponent({
     headingTone(): string {
       return this.image ? 'white' : 'gray-dark'
     },
+    displayFooter(): boolean {
```
edufarre, line 151: Should we add a test to check that this works as expected now?

astyrdoncel: ✅ Test created

TamaraRodriguez (approved):

alex-castelo (approved):

PR #633, opened 2022-12-27
feat: Icon property added to ma-link

catakkio:
### Description ###
- `icon` prop added to `ma-link`. It has the same styles as the `ma-button` component
- Tests created
- New stories to see component behavior with and without icon


src/components/MaLink/MaLink.css
```diff
@@ -121,3 +121,12 @@
   justify-content: center;
   align-items: center;
 }
+
+.ma-link--iconContainer {
```
astyrdoncel, line 125:
```suggestion
.ma-link--icon-container {
```

src/components/MaLink/MaLink.vue
```diff
@@ -69,13 +92,24 @@ export default defineComponent({
         ? `${this.rel || ''} noopener noreferrer`
         : this.rel
     },
+
     getClasses(): { [x: string]: boolean } {
       return {
         [`ma-link--${this.category}`]: !!this.category,
         [`ma-link--${this.size}`]: !!this.size,
         ['ma-link--fullwidth']: !!this.fullWidth,
+        ['ma-link--iconContainer']: !!this.icon,
```
astyrdoncel, line 101:
```suggestion
        ['ma-link--icon-container']: !!this.icon,
```

src/components/MaLink/MaLink.vue
```diff
@@ -1,11 +1,21 @@
 <template>
   <a :class="[getClasses, 'ma-link']" :href="href" :rel="computedRel">
+    <ma-icon
+      v-if="icon"
+      data-testid="icon-link"
+      class="ma-link--icon"
+      :icon="icon"
+      :size="iconSize"
+      :tone="iconTone"
```
astyrdoncel, line 9: I see that this property is not used in my-button, why is it necessary in the link?

src/components/MaLink/MaLink.css
```diff
@@ -121,3 +121,12 @@
...
+  display: inline-flex;
+  align-items: center;
+}
+
+.ma-link--icon {
+  margin-right: 0.5rem;
```
astyrdoncel, line 131:
```suggestion
  margin-right: var(--spacing-xsmall)
```

src/components/MaLink/MaLink.css
```diff
@@ -121,3 +121,12 @@
   justify-content: center;
   align-items: center;
 }
+
+.ma-link--iconContainer {
+  display: inline-flex;
+  align-items: center;
```
astyrdoncel, line 127: Can't this be added by default to the button type? or the default link?

src/components/MaLink/MaLink.vue
icurdu: IMO, this property should be used in both components, or not used in any (adding the icon in the slot and documenting properly the example in the storybook).

icurdu:
Overall, good job adding the prop! The only thing that should be revised is the icon color when hovering over the link category="secondary" (doesn't change from pink to white). I remember discussing this behavior in previous bi-weeklies and I thought it was a bug of ma-icon component, but the link category="no-background-white" changes the color correctly between the interactive states, so I guess it can be fixed.

src/components/MaLink/MaLink.vue
alex-castelo: I think that this should be discussed and could be a BREAKING CHANGE. I suggest to separate the implementation.
First, let's handle the link, and if all works properly, let's do the button.

Anyway, if we don't want to break too many apps, I think `slot`. and `icon` have to live together for a while. What do you think about it.

alex-castelo:
> Overall, good job adding the prop! The only thing that should be revised is the icon color when hovering over the link category="secondary" (doesn't change from pink to white). I remember discussing this behavior in previous bi-weeklies and I thought it was a bug of ma-icon component, but the link category="no-background-white" changes the color correctly between the interactive states, so I guess it can be fixed.

I think the only think remaining is the _icon hover with secondary link_

src/components/MaLink/MaLink.vue
astyrdoncel: @alex-castelo @icurdu
I have already made the change to the `icon` prop for the `ma-link` component.

astyrdoncel:
@icurdu @alex-castelo 
On the other side: **I have seen that there are quite a few inconsistencies with the current `ma-button` styles.**
These changes, being a patch, are in this other PR: link

src/components/MaIcon/index.ts
```diff
@@ -1,3 +1,5 @@
 import MaIcon from './MaIcon.vue'
 
+export * from './MaIcon.types'
+
 export default MaIcon
```
edufarre, line 5: I would not export the types in the same file where we export the component. Because when you would import/use it, It could be misleading. Thus, its import would be:
```ts
// Wherever we would like to use the type
import { Icon } from './MaIcon.types'
```

icurdu (approved):
In the storybook preview, everything looks great :)

src/components/MaIcon/index.ts
astyrdoncel: ✅ Changes made to `ma-link` and `ma-button`

alex-castelo (approved):

edufarre (approved):

PR #634, opened 2023-01-05
fix: Fixed ma-button icon prop styles

astyrdoncel:
### Description ###
- Updated the styles of the `ma-button` to match those of the `ma-link`.
- The associated stories have also been changed to better see the effect of fullWidth on the component.


edufarre (approved):
Beware that you could be having some code issues when trying to merge along with this [PR](https://github.com/holaluz/margarita/pull/633#pullrequestreview-1239999607) as in both Pull Requests you are modifying the MaIcon component

src/components/MaLink/MaLink.spec.ts
```diff
@@ -56,5 +56,51 @@ describe('Link', () => {
 
       expect(getByRole('link')).not.toHaveAttribute('rel')
     })
+
```
edufarre, line 59: 🚀

alex-castelo (approved):
🏃🏽‍♂️

astyrdoncel:
> In the storybook preview, everything looks great! (regarding this PR)
> 
> The only thing I spotted checking all the states is the style of the inline-link (doesn't change its size when selecting large or small prop, image attached), but It may be caused by the storybook configuration and can be done in a different issue.
> 
> image

✅  Well seen! I have already modified it so that you can see if changes are made @icurdu

PR #635, opened 2023-01-11
feat: update button styles

ChaoHu99:
feat: update button styles


edufarre (commented):
why is this a breaking change? it does not change the component's API, thus its behaviour is the expected and it does not change

ChaoHu99:
> why is this a breaking change? it does not change the component's API, thus its behaviour is the expected and it does not change

We thought that it could be a breaking change because it changes the name of the props, but maybe we are wrong. Is there anywhere we can find documentation about how to categorize the PRs?

icurdu:
Everything looks great except for the placement of the loader in the category="tertiary-white"  size="small"
loader-tertiary

edufarre:
> > why is this a breaking change? it does not change the component's API, thus its behaviour is the expected and it does not change
> 
> We thought that it could be a breaking change because it changes the name of the props, but maybe we are wrong. Is there anywhere we can find documentation about how to categorize the PRs?

Yep, you are right, I've missed that, sorry. However, I believe the title of the pr should be `feat: {whatever}` and then the body of the commit have the `BREAKING CHANGE` text

ChaoHu99:
> Everything looks great except for the placement of the loader in the category="tertiary-white" size="small" loader-tertiary
> 
> Update: This position may be fixed in another [PR](https://github.com/holaluz/margarita/pull/633) recently merged

As you said the other PR fixed this issue 
image

icurdu (approved):
LGTM, good job!

sergigraciaj (approved):

PR #637, opened 2023-01-13
feat: Refactor ma stepper 

ferbg:
This is the new Stepper component. The goal of the steppers is to show progress through a sequence of logical, numbered steps vertically and horizontally.

The MaStepper.vue component is the main component where the general styles are defined, passing as props: mode (horizontal/vertical), tone (light/dark) and selected (step indicator). The content must be passed to this component by slot. In order for the content to be displayed correctly MaStepperStep must be used passing the information inside it.

The MaStepperStep.vue component is the step where we define the following props: index(step number), picto, title, description and info. Depending on the needs of use, we could add any content (form, texts, lists, etc.) by means of a slot, substituting-nullifying the mentioned props.

The proposed development idea is to reuse the same component for horizontal (responsive version) and vertical formats.

In the examples stories you can check the caustics contemplated according to design.

You can learn more in the [figma documentation](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=22788%3A9857&t=PGp9okqiNOHxxQSc-1).
image


src/components/MaStepper/MaStepper.vue
```diff
@@ -25,103 +15,59 @@ type StepperTone = 'pink' | 'white'
...
+     * @values horizontal, vertical
      */
-    size: {
-      type: String as PropType<StepperSize>,
-      default: 'large',
-      validator: (val: StepperSize) => ['small', 'large'].includes(val),
-    },
-
+    mode: {
+      type: String,
```
alex-castelo, line 32:
```suggestion
      type: String as PropType<Mode>,
```

src/components/MaStepper/MaStepper.vue
```diff
@@ -25,103 +15,59 @@ type StepperTone = 'pink' | 'white'
...
-    },
-
-    numberTone(): string {
-      return `ma-stepper-${this.tone}`
+    /**
+     * Sets the number and line elements color tone
+     * @values light, dark
+     */
+    tone: {
+      type: String,
```
alex-castelo, line 49:
```suggestion
      type: String as PropType<'light' | 'dark'>,
```

sergigraciaj (approved):
🔝

alex-castelo (approved):

src/components/MaStepper/MaStepper.vue
ferbg: Added `Tone` type and defined as _PropOption_ in the same way as `Mode`
Thanks 😉

PR #638, opened 2023-01-18
MaContainer with full width by default

astyrdoncel:
We have seen that the `ma-container` in projects like `dry-martini` or `maitai` can have a default `width: 100%` because we often use custom css classes to do this.

We also need this functionality for the new `ma-skeleton` component. All parent elements that have this component as a child need to have a defined width, with a max-width.


sergigraciaj (approved):
LGTMIQ+

alex-castelo (approved):

ferbg (approved):

ferbg:
`lumo360` uses ma-container but it is easy to adapt to the change 🚀

carlotamolintr (approved):

PR #651, opened 2023-03-23
feat: Modify link category to be consistent with button

sergigraciaj:
In #635 the button's category are modified to be consistent with Watt, so to be consistent with buttons and Watt here we modify the links category too : D

[Buttons in Watt](https://www.figma.com/file/7RhweKibaUnxX6WYBl7zKW/%E2%9A%A1%EF%B8%8F-Watt-Design-System?node-id=2174%3A66&t=PdNxfhXBfREpNFdB-4)


aquilinopintoa (approved):

edufarre (approved):
I believe this would be a breaking change, right?

